PR #25, opened 2024-08-13
Feature/introduce stripe source

sanjushahgupta:
- added Stripe as an source option in factory.py.
- added StripeAnalyticsSource class in source.py to fetch data from stripe API
- added necessary dependencies to requirements.txt
- added ingestr.src.stripe_analytics as a module in pyproject.toml


ingestr/src/stripe_analytics/README.md
```diff
@@ -0,0 +1,60 @@
+# Stripe
```
karakanb, line 1: we'd like to remove this README file from here.

ingestr/src/.dlt/.sources
```diff
@@ -0,0 +1,28 @@
+engine_version: 1
```
karakanb, line 1: we'd like to remove the whole `.dlt` folder from here as well

sanjushahgupta: ok

ingestr/src/stripe_analytics/README.md
sanjushahgupta: ok

karakanb (approved):

PR #29, opened 2024-08-28
Introduce Airtable  as a source

sanjushahgupta:

- added airtable as a source option in factory.py.
- added airtable source class in source.py to fetch data from the Airtable API.
- added necessary dependencies to requirements.txt.
- added ingestr.src.airtable to pyproject.toml.
- added airtable documentation to docs
- updated README.md to include airtable in the platform sources

Below are the **screenshots** of the destination tables loaded from different source tables.
**Users Table**
users
**Countries Table**
countries
**Users Table + Countries Table**
users_countries


docs/supported-sources/airtable.md
```diff
@@ -0,0 +1,34 @@
...
+```plaintext
+ airtable://?access_token=<access_token>&base_id=<base_id>
+```
+
+URI parameters:
+
+- `base_id`: A unique identifier for an Airtable base.
+- `access_token`: A personal access token for authentication with the Airtable API.
+
+The URI is used to connect to the Airtable API for extracting data. More details on setting up Airtable integrations can be found.
```
karakanb, line 20: "More details on setting up Airtable integrations can be found." where?

ingestr/src/sources.py
```diff
@@ -480,3 +481,35 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+    def handles_incrementality(self) -> bool:
+        return True
+
+    # airtable://?access_token=<access_token>&base_id=<base_id>
+
+    def dlt_source(self, uri: str, table: str, **kwargs):
+        if kwargs.get("incremental_key"):
+            raise ValueError(
+                "Airtable takes care of incrementality on its own, you should not provide incremental_key"
+            )
```
karakanb, line 496: is that actually the case? it seems like Airtable does not support incremental loading, therefore we should return an error that says incremental loading is not supported for Airbyte.

Please also document the same in the docs, similar to how Google Sheets source is documented.

karakanb (approved):

docs/supported-sources/airtable.md
sanjushahgupta: I have added missing url. 

Screenshot 2024-08-29 at 13 31 20

ingestr/src/sources.py
sanjushahgupta: Yes, it doesn’t support incremental loading. I have made necessaries changes.


Screenshot 2024-08-29 at 13 00 16
Screenshot 2024-08-29 at 13 09 18

PR #37, opened 2024-09-10
Introduce Adjust as new source

sanjushahgupta:
Below are changes made in this PR:
- added adjust as a new source
- added campaigns as a resource

### DIFF
--- a/ingestr/src/adjust/_init_.py
+++ b/ingestr/src/adjust/_init_.py
@@ -5,18 +5,19 @@ from dlt.sources import DltResource
 
 from .helpers import AdjustAPI
 
+
 @dlt.source(max_table_nesting=0)
 def adjust_source(
     start_date: str,
     end_date: str,
     api_key: str,
 ) -> Sequence[DltResource]:
-
     @dlt.resource(write_disposition="merge", merge_key="day")
     def campaigns():
         adjust_api = AdjustAPI(api_key=api_key)
         yield from adjust_api.fetch_report_data(
-            start_date = start_date,
-            end_date = end_date,
+            start_date=start_date,
+            end_date=end_date,
         )
+
     return campaigns
--- a/ingestr/src/adjust/_init_.py
+++ b/ingestr/src/adjust/_init_.py
@@ -1,12 +1,9 @@
 from typing import Sequence
 
 import dlt
-import pendulum
-from dlt.common.time import ensure_pendulum_datetime
 from dlt.sources import DltResource
 
 from .helpers import AdjustAPI
-
 
 @dlt.source(max_table_nesting=0)
 def adjust_source(
@@ -14,20 +11,12 @@ def adjust_source(
     end_date: str,
     api_key: str,
 ) -> Sequence[DltResource]:
-    start_date_obj = ensure_pendulum_datetime(start_date)
 
     @dlt.resource(write_disposition="merge", merge_key="day")
-    def campaigns(
-        start_date=dlt.sources.incremental("day", start_date_obj.isoformat()),
-    ):
-        print("start_date_incremental", start_date.start_value)
-        formatted_start_date = pendulum.parse(start_date.start_value).format(
-            "YYYY-MM-DD"
+    def campaigns():
+        adjust_api = AdjustAPI(api_key=api_key)
+        yield from adjust_api.fetch_report_data(
+            start_date = start_date,
+            end_date = end_date,
         )
-
-        adjust_api = AdjustAPI(
-            start_date=formatted_start_date, end_date=end_date, api_key=api_key
-        )
-        yield from adjust_api.fetch_report_data()
-
     return campaigns
--- a/ingestr/src/adjust/_init_.py
+++ b/ingestr/src/adjust/_init_.py
@@ -1,0 +1,33 @@
+from typing import Sequence
+
+import dlt
+import pendulum
+from dlt.common.time import ensure_pendulum_datetime
+from dlt.sources import DltResource
+
+from .helpers import AdjustAPI
+
+
+@dlt.source(max_table_nesting=0)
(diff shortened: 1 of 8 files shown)


ingestr/src/adjust/_init_.py
```diff
@@ -0,0 +1,33 @@
...
+    end_date: str,
+    api_key: str,
+) -> Sequence[DltResource]:
+    start_date_obj = ensure_pendulum_datetime(start_date)
+
+    @dlt.resource(write_disposition="merge", merge_key="day")
+    def campaigns(
+        start_date=dlt.sources.incremental("day", start_date_obj.isoformat()),
+    ):
+        print("start_date_incremental", start_date.start_value)
```
karakanb, line 23: do not forget to clean this up please

ingestr/src/adjust/_init_.py
```diff
@@ -0,0 +1,33 @@
...
+        formatted_start_date = pendulum.parse(start_date.start_value).format(
+            "YYYY-MM-DD"
+        )
+
+        adjust_api = AdjustAPI(
+            start_date=formatted_start_date, end_date=end_date, api_key=api_key
+        )
+        yield from adjust_api.fetch_report_data()
```
karakanb, line 31: you should give the api_key to the class, and the dates to the function, simply because api_key is common for all functions, whereas dates don't have to be.

```suggestion
        adjust_api = AdjustAPI(api_key=api_key)
        yield from adjust_api.fetch_report_data(start_date=formatted_start_date, end_date=end_date)
```

ingestr/src/adjust/helpers.py
```diff
@@ -0,0 +1,60 @@
...
+        self.end_date = end_date
+        self.api_key = api_key
+        self.default_dimensions = "campaign,day,app,store_type,channel,country"
+        self.default_metrics = (
+            "network_cost,all_revenue_total_d0,ad_revenue_total_d0,revenue_total_d0,"
+            "all_revenue_total_d1,ad_revenue_total_d1,revenue_total_d1,all_revenue_total_d3,"
+            "ad_revenue_total_d3,revenue_total_d3,all_revenue_total_d7,ad_revenue_total_d7,"
+            "revenue_total_d7,all_revenue_total_d14,ad_revenue_total_d14,revenue_total_d14,"
+            "all_revenue_total_d21"
+        )
```
karakanb, line 18: do not make them class attributes, instead make them variables outside of the class, and then accept metrics optionally from different functions.

e.g. the following code should work:
```python
client = AdjustAPI(api_key="my-api-key")
client.fetch_report_data(dimensions=['campaign', 'day'], metrics=['network_cost', 'revenue_total_d1'])
```

in order to get this to work, you need to change the functions.

ingestr/src/adjust/helpers.py
```diff
@@ -0,0 +1,60 @@
...
+        self.uri = "https://automate.adjust.com/reports-service/report"
+
+    def fetch_report_data(self):
```
karakanb, line 21:
```suggestion
    def fetch_report_data(self, metrics=DEFAULT_METRICS, dimensions=DEFAULT_DIMENSIONS):
```

ingestr/src/adjust/helpers.py
```diff
@@ -0,0 +1,60 @@
...
+            "date_period": f"{self.start_date}:{self.end_date}",
+            "dimensions": self.default_dimensions,
+            "metrics": self.default_metrics,
+            "utc_offset": "+00:00",
+            "ad_spend_mode": "network",
+            "attribution_source": "first",
+            "attribution_type": "all",
+            "cohort_maturity": "immature",
+            "reattributed": "all",
+            "sandbox": "false",
```
karakanb, line 33: these should be optionally configurable as well

ingestr/src/adjust/helpers.py
```diff
@@ -0,0 +1,60 @@
...
+        ).session
+
+        response = request_client.get(self.uri, headers=headers, params=params)
+        if response.status_code == 200:
+            result = response.json()
+            items = result.get("rows", [])
+
+            for item in items:
+                response_days = pendulum.parse(item["day"])
+                incremntal_days = pendulum.parse(self.start_date)
```
karakanb, line 56: do not do this for every item, move this out of the loop so that you parse it just once

ingestr/src/adjust/helpers.py
```diff
@@ -0,0 +1,60 @@
...
+            request_max_attempts=12,
+            request_backoff_factor=2,
+        ).session
+
+        response = request_client.get(self.uri, headers=headers, params=params)
+        if response.status_code == 200:
+            result = response.json()
+            items = result.get("rows", [])
+
+            for item in items:
```
karakanb, line 54: are you sure the API response doesn't return multiple pages?

ingestr/src/sources.py
```diff
@@ -652,3 +653,36 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+
+class AdjustSource:
+    def handles_incrementality(self) -> bool:
+        return False
+
+    def dlt_source(self, uri: str, table: str, **kwargs):
+        if kwargs.get("incremental_key"):
+            raise ValueError("Incremental loads are not supported for Adjust")
+
+        api_key = None
```
karakanb, line 666: you don't need this line

ingestr/src/adjust/helpers.py
sanjushahgupta: According to the documentation, it says pagination: null 
In my response I am not getting pagination key.  So I think API does not return multiple pages.
Screenshot 2024-09-11 at 12 25 45
Screenshot 2024-09-11 at 12 27 07

karakanb (approved):

PR #38, opened 2024-09-16
implemented passing column hints to the resource(s) dynamically

canerdev:


ingestr/src/factory_test.py
```diff
@@ -11,3 +11,88 @@ def test_scheme_is_parsed_from_uri_correctly():
...
+def test_parse_columns_all_data_types():
+    input_columns = [
+        "col1:text", "col2:double", "col3:bool", "col4:timestamp",
+        "col5:bigint", "col6:binary", "col7:complex", "col8:decimal",
+        "col9:wei", "col10:date", "col11:time"
+    ]
+    result = parse_columns(input_columns)
+    assert len(result) == 11
+    for col, props in result.items():
+        assert props["nullable"] == False
```
karakanb, line 44: let's test for the array you expect here right away, we should expect the value to be exactly equal to what you expect.

ingestr/src/factory_test.py
```diff
@@ -11,3 +11,88 @@ def test_scheme_is_parsed_from_uri_correctly():
...
+    with pytest.raises(ValueError, match="Invalid column name"):
+        parse_columns(["invalid@column:text"])
+
+def test_parse_columns_invalid_data_type():
+    with pytest.raises(ValueError, match="Invalid data type"):
+        parse_columns(["valid_name:invalid_type"])
+
+def test_parse_columns_invalid_nullable():
+    with pytest.raises(ValueError, match="Argument format is incorrect"):
+        parse_columns(["valid_name:text:invalid"])
```
karakanb, line 66: seems like you could merge all of these in a single test?

ingestr/src/factory_test.py
```diff
@@ -11,3 +11,88 @@ def test_scheme_is_parsed_from_uri_correctly():
...
+    ]
+    with pytest.raises(ValueError):
+        parse_columns(input_columns)
+
+def test_parse_columns_case_sensitivity():
+    input_columns = ["COL1:TEXT", "Col2:Double:NULLABLE"]
+    result = parse_columns(input_columns)
+    assert "COL1" in result
+    assert result["COL1"]["data_type"] == "text"
+    assert result["Col2"]["nullable"] == True
```
karakanb, line 83: test with an exact array, sth like this:
```suggestion
    expected = [
        {
            "name": "COL1",
            "data_type": "text",
            "nullable": False
        },
        {
            "name": "Col2",
            "data_type": "double",
            "nullable": True
        }
    ]
    result = parse_columns(input_columns)
    assert result == expected
```

ingestr/src/factory.py
```diff
@@ -152,3 +153,45 @@ def get_destination(self) -> DestinationProtocol:
...
+
+            column_name, data_type, nullability = validate_properties(properties)
+
+            column_properties = {
+                "data_type": data_type,
+                "nullable": nullability
+            }
+            cols[column_name] = column_properties
+
+    return cols
```
karakanb, line 197: you are missing a newline at the end of the file

PR #44, opened 2024-10-16
Feat/appsflyer

sanjushahgupta:
The following changes are made in this PR:
- date partitioning to fetch all requested data
- handle end date for cohort metrics
- use install time as the incremental key


ingestr/src/appsflyer/_init_.py
```diff
@@ -11,14 +13,38 @@
 def appsflyer_source(
     api_key: str, start_date: str, end_date: str
 ) -> Iterable[DltResource]:
+    start_date_obj = ensure_pendulum_datetime(start_date)
     client = AppsflyerClient(api_key)
 
-    @dlt.resource(write_disposition="merge", merge_key="install_time")
-    def campaigns() -> Iterable[TDataItem]:
-        yield from client.fetch_campaigns(start_date, end_date)
+    @dlt.resource(write_disposition="merge", merge_key="Install Time")
```
karakanb, line 19: the key is supposed to be as the fields in JSON, why did you go with `Install Time` instead of the previous version?

ingestr/src/appsflyer/_init_.py
```diff
@@ -11,14 +13,38 @@
...
+    def campaigns(
+        updated=dlt.sources.incremental('["Install Time"]', start_date_obj.isoformat()),
+    ) -> Iterable[TDataItem]:
+        current_start_time = datetime.fromisoformat(updated.start_value).date()
```
karakanb, line 23: shouldn't this be `updated.last_value`?

ingestr/src/appsflyer/_init_.py
```diff
@@ -11,14 +13,38 @@
...
+        end_date_time = datetime.fromisoformat(end_date).date()
+        
+        while current_start_time < end_date_time:
+            current_end_time = current_start_time + timedelta(days=30)
+            next_end_date = min(current_end_time, end_date_time)
+            yield from client.fetch_campaigns(
+                start_date=current_start_time.isoformat(), end_date=next_end_date.isoformat()
+            )
+            print(current_start_time, next_end_date)
```
karakanb, line 32: we shouldn't print this

ingestr/src/appsflyer/_init_.py
```diff
@@ -11,14 +13,38 @@
...
-    def campaigns() -> Iterable[TDataItem]:
-        yield from client.fetch_campaigns(start_date, end_date)
+    @dlt.resource(write_disposition="merge", merge_key="Install Time")
+    def campaigns(
+        updated=dlt.sources.incremental('["Install Time"]', start_date_obj.isoformat()),
+    ) -> Iterable[TDataItem]:
+        current_start_time = datetime.fromisoformat(updated.start_value).date()
+        end_date_time = datetime.fromisoformat(end_date).date()
+        
+        while current_start_time < end_date_time:
```
karakanb, line 26: I'd move this logic into the internal appsflyer client so that we don't have to repeat it

ingestr/src/appsflyer/client.py
```diff
@@ -89,13 +90,12 @@ def fetch_campaigns(
...
-            "cohort_day_14_revenue_per_user",
-            "cohort_day_21_total_revenue_per_user",
-            "cohort_day_21_revenue_per_user",
             "retention_day_7",
         ]
-        return self._fetch_data(start_date, end_date, metrics=metrics)
+        max_cohort_duration = 7
+        max_allowed_end_date = (datetime.now() - timedelta(days=max_cohort_duration)).strftime('%Y-%m-%d')
+        adjusted_end_date = min(end_date, max_allowed_end_date)
+        return self._fetch_data(start_date, adjusted_end_date, metrics=metrics)
```
karakanb, line 98: why?

ingestr/src/appsflyer/_init_.py
sanjushahgupta: Because "Install Time" is the JSON key not "install_time". My understanding was incorrect.

ingestr/src/appsflyer/_init_.py
sanjushahgupta: While testing, both start_value and last_value were giving the same value, so I used start_value.

ingestr/src/appsflyer/_init_.py
sanjushahgupta: Done

ingestr/src/appsflyer/_init_.py
sanjushahgupta: Done

ingestr/src/appsflyer/client.py
sanjushahgupta: We can only fetch cohort metric data from more than 7 days ago. If a user tries to fetch cohort data for today, Appsflyer shows an error "429." To handle this, we limit the end date to 7 days before the current date. For example, if a user is trying to fetch data from 2024-09-10 to 2024-10-20, the data will be fetched from 2024-09-10 to 2024-10-13 instead. Otherwise, it will show an error, and we won't be able to fetch data (this issue occurs only in the campaigns resource).

sanjushahgupta:
Noted, I will work on these in the new PR.

PR #68, opened 2024-12-10
Feat/tiktok

sanjushahgupta:
- Added custom_report resource for the TikTok source
- Users can fetch custom reports based on dimensions and metrics
- Incremental loading is supported if the user provides the stat_time dimension
- Pagination is supported

Note:
Filter logic is not yet implemented and will be added soon


ingestr/src/tiktok_ads/tiktok_helpers.py
```diff
@@ -0,0 +1,125 @@
...
+        start_time = ensure_pendulum_datetime(start_time).to_date_string()
+        end_time = ensure_pendulum_datetime(end_time).to_date_string()
+
+        self.params = {
+            "advertiser_id": advertiser_id,
+            "report_type": "BASIC",
+            "data_level": data_level,
+            "start_date": start_time,
+            "end_date": end_time,
+            "page_size": 1000,
```
karakanb, line 84: is this configurable? ideally we should get it from the `--page-size` variable in ingestr

ingestr/src/tiktok_ads/tiktok_helpers.py
```diff
@@ -0,0 +1,125 @@
...
+            "dimensions": json.dumps(dimensions),
+            "metrics": json.dumps(metrics),
+        }
+
+        while True:
+            self.params["page"] = current_page
+            response = create_client().get(
```
karakanb, line 91: why create the client in every step of the loop?

ingestr/src/tiktok_ads/tiktok_helpers.py
```diff
@@ -0,0 +1,125 @@
...
+            result = response.json()
+            if result.get("message") != "OK":
+                raise ValueError(result.get("message", ""))
+
+            result_data = result.get("data", {})
+            items = result_data.get("list", [])
+
+            flat_structure(items=items, time_zone=self.time_zone)
+
+            all_items.extend(items)
```
karakanb, line 104: I suggest yielding these pages instead of collecting them, otherwise we might run into memory issues

ingestr/src/sources.py
```diff
@@ -37,6 +38,7 @@
 from ingestr.src.slack import slack_source
 from ingestr.src.stripe_analytics import stripe_source
 from ingestr.src.table_definition import table_string_to_dataclass
+from ingestr.src.tiktok_ads._init_ import tiktok_source
```
turtleDev, line 41: If you change `ingestr/src/tiktok_ads/_init_.py` to `ingestr/src/tiktok_ads/__init__.py`, you will be able to treat `tiktok_ads` as a python module.

That way, the import statement can change to:
```python
from ingestr.src.tiktok_ads import tiktok_source
```

ingestr/src/sources.py
```diff
@@ -997,6 +997,77 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+class TikTokSource:
+    # tittok://?access_token=<access_token>&advertiser_id=<advertiser_id>
+    def handles_incrementality(self) -> bool:
+        return True
+
+    def dlt_source(self, uri: str, table: str, **kwargs):
+        endpoint = "custom_reports"
+
+        parsed_uri = urlparse(uri)
+        source_fields = parse_qs(parsed_uri.query)
```
turtleDev, line 1009: Can tiktok access token contain `=`, `+` or `&` ? We may need to escape those.

ingestr/src/sources.py
```diff
@@ -997,6 +997,77 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+        advertiser_id = source_fields.get("advertiser_id")
+        if not advertiser_id:
+            raise ValueError("advertiser_id is required to connect to TikTok")
+
+        if kwargs.get("interval_start"):
+            start_date = ensure_pendulum_datetime(
+                str(kwargs.get("interval_start"))
+            ).in_tz(time_zone[0])
+        else:
+            Default_date = pendulum.now().subtract(days=90)
```
turtleDev, line 1026: Do we need a default start date in order to be able to load data?

ingestr/src/tiktok_ads/tiktok_helpers.py
sanjushahgupta: We don't need to create a client every time. It's my mistake. I have corrected that.

ingestr/src/sources.py
sanjushahgupta: Yes, we need it in this case because the API expects a start and end date.

ingestr/src/sources.py
sanjushahgupta: Regarding this I am not sure. We have not escaped them before, so we need to discuss this with the team.

karakanb: I agree with leaving this as is.

karakanb (approved):

PR #69, opened 2024-12-11
Add Asana source

turtleDev:
**Note**
I didn't add any tests, because there are no mock servers for Asana API.


ingestr/src/asana_source/__init__.py
```diff
@@ -0,0 +1,264 @@
+"""
```
albertobruin, line 1: I see that we dont add `_source` to the package for other platforms, even when they are only source. Maybe we keep that for consistency?

requirements.txt
```diff
@@ -29,4 +29,5 @@ pyairtable==2.3.3
 types-requests==2.32.0.20240907
 s3fs==2024.10.0
 pyarrow==18.1.0
-pyathena==3.9.0
\ No newline at end of file
+pyathena==3.9.0
+asana<5.0.0
```
albertobruin, line 33: would this mean a major 6 could be installed exposing us to breaking changes? I'd rather lock it to a specific one version

ingestr/src/asana_source/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+    TAG_FIELDS,
+    TASK_FIELDS,
+    TEAMS_FIELD,
+    USER_FIELDS,
+    WORKSPACE_FIELDS,
+)
+
+
+@dlt.source
+def asana_source() -> Any:  # should be Sequence[DltResource]:
```
albertobruin, line 31: I see that all other sources return some sort of iterable or list, why not do the same? Also I see that many have the `start_date` and `end_date` arguments to the `*_source` function. Why not here?

requirements.txt
turtleDev: fixed!

I've pinned the version to the highest matching package version.

ingestr/src/asana_source/__init__.py
turtleDev: naming the package to `asana` causes ambiguity in imports when importing the `asana` client.

This was originally called `asana` package, but when importing `asana` client, it would cause a circular import.

dlt also ran into the same issue, and as a consequence, their version is called `asana_dlt`

ingestr/src/sources.py
```diff
@@ -994,3 +995,42 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+        "sections",
+        "tags",
+        "tasks",
+        "stories",
+        "teams",
+        "users",
+    ]
+
+    def handles_incrementality(self) -> bool:
+        return False
```
karakanb, line 1013: it does handle incrementality, in the sense that the user cannot dictate it from the outside with flags like` --incremental-strategy`, even though the source says "we don't support incrementality".

PR #70, opened 2024-12-11
Add google analytics as source

sanjushahgupta:


ingestr/src/google_analytics/setup_script_gcp_oauth.py
```diff
@@ -0,0 +1,55 @@
+"""
```
karakanb, line 1: I suggest removing this script.

docs/supported-sources/google_analytics.md
```diff
@@ -0,0 +1,44 @@
...
+## URI format
+The URI format for Google Analytics is as follows:
+
+```plaintext
+googleanalytics://?credentials_path=/path/to/service/account.json&property_id=<property_id>
+```
+
+URI parameters:
+- `credentials_path` = The path to the service account JSON file.
+- `property_id` = It is a unique number that identifies a particular property. You will need to explicitly pass it to get data from the property that you're interested in. For example, if the property that you want to get data from is “GTA-Google Shop”, then you will need to pass its property id.
```
karakanb, line 15:
```suggestion
- `credentials_path`: The path to the service account JSON file.
- `property_id`: It is a unique number that identifies a particular property on Google Analytics. [Follow this guide](https://developers.google.com/analytics/devguides/reporting/data/v1/property-id#what_is_my_property_id) if you don't know your property ID.
```

requirements.txt
```diff
@@ -30,4 +30,7 @@ types-requests==2.32.0.20240907
 s3fs==2024.10.0
 pyarrow==18.1.0
 pyathena==3.9.0
-asana==3.2.3
\ No newline at end of file
+google-analytics-data==0.18.15
+google-auth-oauthlib==1.2.1
```
karakanb, line 34: do we need this oauth lib since we use service accounts? can you please test this after removing this library?

sanjushahgupta: Yeah, We don't need this. It's working fine after removing these library.

PR #71, opened 2024-12-16
Feat: add DynamoDB Source

turtleDev:
* Adds dynamodb source
* fixes credentials escapes during url parsing in s3
* refactors source and destinations lists in `SourceDestinationFactory`


ingestr/src/sources.py
```diff
@@ -1034,3 +1037,64 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+        )
+
+        incremental = None
+        incremental_key = kwargs.get("incremental_key", "").strip()
+
+        if incremental_key:
+            incremental = dlt.sources.incremental(
+                incremental_key,
+                initial_value=isotime(kwargs.get("interval_start")),
+                end_value=isotime(kwargs.get("interval_end")),
```
karakanb, line 1097: I don't think we should convert them to iso formats, it could be that the user is filtering for a date field instead. I'd also remove this helper function since it wouldn't be used.

docs/supported-sources/dynamodb.md
```diff
@@ -0,0 +1,40 @@
...
+- `secret_access_key`: Password for the IAM account.
+
+
+## Setting up a DynamoDB integration
+
+### Prerequisites
+* AWS IAM access key pair.
+* A DynamoDB Table that you will to load data from
+
+To obtain the access keys, use the IAM console on AWS. See [IAM Documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html) for more information.
```
karakanb, line 24: are we able to share a bit more on the particular IAM permissions that are needed for this?

docs/supported-sources/dynamodb.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Example
```
karakanb, line 26: can you please add an incremental example?

ingestr/main_test.py
```diff
@@ -1538,3 +1545,188 @@ def test_db_to_db_exclude_columns(source, dest):
...
+            
+        dest.stop()
+
+    def append_test(dest, dynamodb):
+        dest_uri = dest.start()
+        dest_table = f"public.dynamodb_{get_random_string(5)}"
+
+        # connection pooling causes issues with duckdb, when the connection
+        # is reused below, so we disable pooling.
+        dest_engine = sqlalchemy.create_engine(dest_uri, poolclass=NullPool)
```
karakanb, line 1656: this is pretty clever, I had to do an uglier solution before, didn't know I could do this 👍

ingestr/main_test.py
```diff
@@ -1538,3 +1545,188 @@ def test_db_to_db_exclude_columns(source, dest):
...
+        )
+        assert_success(result)
+        rows = dest_engine.execute(f"SELECT id, updated_at from {dest_table} ORDER BY id").fetchall()
+        assert len(rows) == 2
+        for i in range(len(rows)):
+            assert rows[i][0] == dynamodb.data[i]["id"]
+            assert rows[i][1] == pendulum.parse(dynamodb.data[i]["updated_at"])
+
+        # ingest the rest
+        result = invoke_ingest_command(
```
karakanb, line 1701: why not run another attempt once more, similar to the above, to check the data? also, I'd suggest having some tests for other strategies like append or delete insert.

karakanb (approved):

PR #89, opened 2025-01-23
Integrate ClickHouse as a destination and add integration testing

sanjushahgupta:
The following changes are made in this PR:
- added ClickHouse as a new destination.
- added documentation for it.
- integration testing for clickHouse as a destination

The destination uri looks like:
clickhouse://<username>:<password>@<host>:<port>?http_port=<http_port>


ingestr/src/destinations.py
```diff
@@ -261,3 +264,70 @@ def dlt_run_params(self, uri: str, table: str, **kwargs) -> dict:
...
+        if not port:
+            raise ValueError(
+                "The TCP port of the ClickHouse server is required to establish a connection."
+            )
+
+        http_port = (
+            int(parsed_uri.query.split("http_port=")[1])
+            if "http_port=" in parsed_uri.query
+            else 8123
+        )
```
karakanb, line 308: this logic is wrong. I think this would fail if the uri was like:
```
clickhouse://user:pw@host:port?http_port=1234&someotherkey=someothervalue
```

which would try to get `1234&someotherkey=someothervalue` and try to cast it to integer.

do it like this instead:
```suggestion
        http_port = (
            int(urlparse.parse_qs(parsed_uri.query)['http_port'][0])
            if "http_port=" in parsed_uri.query
            else 8123
        )
```

requirements.txt
```diff
@@ -43,4 +43,7 @@ asana==3.2.3
 flatten_json==0.1.14
 dataclasses-json==0.6.7
 gcsfs==2024.10.0
+clickhouse-connect == 0.8.14
+clickhouse-driver ==  0.2.9
+clickhouse-sqlalchemy<0.3.0
```
karakanb, line 48: pin this to an exact version please

requirements.txt
```diff
@@ -43,4 +43,7 @@ asana==3.2.3
 flatten_json==0.1.14
 dataclasses-json==0.6.7
 gcsfs==2024.10.0
+clickhouse-connect == 0.8.14
+clickhouse-driver ==  0.2.9
```
karakanb, line 47:
```suggestion
clickhouse-connect==0.8.14
clickhouse-driver==0.2.9
```

karakanb (approved):

PR #103, opened 2025-01-31
Clickhouse: support both secure and non-secure connection

sanjushahgupta:


ingestr/src/destinations.py
```diff
@@ -302,9 +302,16 @@ def dlt_dest(self, uri: str, **kwargs):
 
         query_params = parse_qs(parsed_uri.query)
         http_port = (
-            int(query_params["http_port"][0]) if "http_port" in query_params else 8123
+            int(query_params["http_port"][0]) if "http_port" in query_params else 8443
```
turtleDev, line 305: We can set the default value by referencing `secure`.

If `secure` is 1 then 8443 else 8123.

turtleDev (approved):

karakanb (approved):

PR #105, opened 2025-02-04
add clickhouse as a source

sanjushahgupta:
Users can load data from a ClickHouse database to a  desired destination.
- source_uri = clickhouse://username:password@host:port?secure
 - By default, secure=1.
 
Screenshot 2025-02-04 at 14 03 19

### DIFF
--- a/ingestr/src/sources.py
+++ b/ingestr/src/sources.py
@@ -131,6 +131,11 @@ class SqlSource:
 
         if uri.startswith("mysql://"):
             uri = uri.replace("mysql://", "mysql+pymysql://")
+        
+        if uri.startswith("clickhouse://"):
+            uri = uri.replace("clickhouse://", "clickhouse+native://")
+            if "secure=" not in uri:
+              uri += "?secure=1"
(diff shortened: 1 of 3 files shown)


ingestr/src/sources.py
```diff
@@ -131,6 +131,11 @@ def dlt_source(self, uri: str, table: str, **kwargs):
 
         if uri.startswith("mysql://"):
             uri = uri.replace("mysql://", "mysql+pymysql://")
+        
+        if uri.startswith("clickhouse://"):
+            uri = uri.replace("clickhouse://", "clickhouse+native://")
+            if "secure=" not in uri:
+              uri += "?secure=1"
```
turtleDev, line 138: This won't work in all cases. For example if the user specifies the uri as 

`clickhouse://user:pass@host?http=port`

It'll add another `?` at the end, which isn't a valid URI

turtleDev (changes requested):

turtleDev (approved):

PR #127, opened 2025-02-16
clickhouse: refactor source uri

sanjushahgupta:
Refactored the ClickHouse source uri to properly handle the secure and http_port param, ensuring that the same uri can be used for source and destination. 

source

dest

ingestion


ingestr/src/sources.py
```diff
@@ -136,10 +136,47 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+
+            query_params = parse_qs(parsed_uri.query)
+
+            if "http_port" in query_params:
+                del query_params["http_port"]
+
+            if "secure" not in query_params:
+                query_params["secure"] = ["1"]
+            else:
+                query_params["secure"] = [query_params["secure"][0]]
```
turtleDev, line 175: Isn't this redundant? If "secure" exists, we don't need to recreate it

ingestr/src/sources.py
```diff
@@ -136,10 +136,47 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+
+            uri = parsed_uri._replace(
```
turtleDev, line 177: We shouldn't rely on non-public APIs. Use `urlunparse` instead.

ingestr/src/sources.py
```diff
@@ -15,7 +15,7 @@
     Optional,
     Union,
 )
-from urllib.parse import ParseResult, parse_qs, quote, urlparse
+from urllib.parse import ParseResult, parse_qs, quote, urlencode, urlparse, urlunparse
```
turtleDev, line 19:
```suggestion
import copy
```

ingestr/src/sources.py
```diff
@@ -136,11 +136,52 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+            uri = urlunparse(
+                (
+                    "clickhouse+native",
+                    parsed_uri.netloc,
+                    "",
+                    "",
+                    urlencode(query_params, doseq=True),
+                    "",
+                )
+            )
```
turtleDev, line 184:
```suggestion
            parsed_uri = copy.replace(parsed_uri, scheme="clickhouse+native", query=urlencode(query_params, doseq=True)
            uri = urlunparse(*parsed_uri)
```

turtleDev (approved):

PR #141, opened 2025-03-04
lock packages version

albertobruin:


turtleDev (dismissed):

karakanb:
are we not gonna do the `requirements.in` and then `requirements.txt` trick that we spoke about? example here: https://docs.astral.sh/uv/pip/compile/#locking-requirements

PR #166, opened 2025-04-15
Frankfurter update

jbrand-dsp:
Adding the following for Frankfurter:
- updated validation checks
- integration tests 
- updated documentation


docs/supported-sources/frankfurter.md
```diff
@@ -41,7 +41,9 @@ ingestr ingest \
 ### **`--interval-end` (Optional)**
 - **Description**: The end date for fetching historical exchange rates.
 - **Value**: A date in the format `YYYY-MM-DD` (e.g., `'2025-03-28'`).
-- **Purpose**: Defines the ending point for fetching historical data. If not provided, it defaults to the value of `--interval-start`.
+- **Purpose**: Defines the ending point for fetching historical data.
```
turtleDev, line 44:
```suggestion
- **Purpose**: Defines the end point for fetching historical data. 
```

ingestr/src/frankfurter/helpers.py
```diff
@@ -19,6 +19,11 @@ def get_path_with_retry(path: str) -> StrAny:
 def validate_dates(start_date: datetime, end_date: datetime) -> None:
     current_date = pendulum.now()
 
+    print("START DATE")
```
turtleDev, line 22: we can remove these debug print statements

ingestr/main_test.py
```diff
@@ -2953,6 +2953,119 @@ def test_applovin_source(testcase):
...
+        schema = f"testschema_frankfurter_{get_random_string(5)}"
+        dest_table = f"{schema}.frankfurter_{get_random_string(5)}"
+        result = invoke_ingest_command(
+            "frankfurter://",
+            "invalid table",
+            dest_uri,
+            dest_table,
+        )
+        assert result.exit_code != 0
+        assert has_exception(result.exception, ValueError)
```
turtleDev, line 2967: This should be `ingestr.src.errors.UnsupportedResourceError`

ingestr/main_test.py
```diff
@@ -2953,6 +2953,119 @@ def test_applovin_source(testcase):
...
+        start_date = pendulum.now().subtract(days=1).format("YYYY-MM-DD")
+        end_date = pendulum.now().add(days=1).format("YYYY-MM-DD")
+        result = invoke_ingest_command(
+            "frankfurter://",
+            "exchange_rates",
+            dest_uri,
+            dest_table,
+            interval_start=start_date,
+            interval_end=end_date,
+        )
```
turtleDev, line 3025: Although I like the idea of these tests, it can cause subtle issues for our users.

For instance, depending on which timezone frankfurter uses, the `interval_end` parameter may be valid.

turtleDev: solutions:
1. Remove --interval-end checks exceeding the current date. Users are responsible for providing the correct parameters.
    -- or --
2. Figure out what timezone frankfurter uses and translate all date/time inputs to that timezone and then run validation on end date.

docs/supported-sources/frankfurter.md
```diff
@@ -41,7 +41,9 @@ ingestr ingest \
 ### **`--interval-end` (Optional)**
 - **Description**: The end date for fetching historical exchange rates.
 - **Value**: A date in the format `YYYY-MM-DD` (e.g., `'2025-03-28'`).
-- **Purpose**: Defines the ending point for fetching historical data. If not provided, it defaults to the value of `--interval-start`.
+- **Purpose**: Defines the ending point for fetching historical data. 
+    - If not provided, it defaults to the value of `--interval-start`.
```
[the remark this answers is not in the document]
jbrand-dsp: Yeah, this was due to a design decision. Sabri suggested having a single historical date as part of exchange rates, as one is a subset of the other. I therefore diverged slightly from the API.

The easiest solution to this problem would be to require an end date or throw an exception.

A more involved solution would be to create follow the API and default to latest and to implement a table for a single historical date and call it something like historical_date. It would probably be a more elegant solution. I started the implementation and then realised I might just be creating more review work for you and more testing work for me.

What do you think?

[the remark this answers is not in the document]
turtleDev: I see where you are coming from.

I can think of a few scenario's from a user's prespective:
1. they need the latest exchange rate to covert currencies for online-transactions.
2. they want to keep a historical record of currency rates 
 
 In case (1) the user will anyway use the `latest` resource. For (2) the user would probably want to have upto-date data from a specific starting point. 
 
 Can you think of any scenario where the user would want only a certain day's exchange rate? And if so, should that be the default expectation with everyone?

ingestr/src/frankfurter/__init__.py
```diff
@@ -94,25 +94,34 @@ def latest() -> Iterator[dict]:
...
 ) -> Iterator[dict]:
     """
     Fetches exchange rates for a specified date range.
-    If only start_date is provided, fetches data for that date.
+    If only start_date is provided, fetches data until now.
     If both start_date and end_date are provided, fetches data for each day in the range.
     """
-    start_date_str = ensure_pendulum_datetime(start_date).format("YYYY-MM-DD")
-    end_date_str = ensure_pendulum_datetime(end_date).format("YYYY-MM-DD")
+    # Ensure start_date.last_value is not None
```
jbrand-dsp, line 114: This was formatting to past the make lint tests. start_date.last_value could in fact not be since start date defaults to today's date if no start_date is given.

turtleDev: yeah I've experienced similar issues especially with union types. You can just add a `# type: ignore` comment to disable the linter for that line.

turtleDev:
Bug:
```
$ ingestr ingest --source-uri 'frankfurter://' --source-table 'currencies' --dest-uri 'duckdb://out.db' --dest-table 'public.currencies'
$ duckdb out.db '.tables'
_dlt_loads
_dlt_version
_dlt_pipeline_state
unknown_report
```

For some reason it names the table as `unknown_report` instead of `public.currencies`

ingestr/src/frankfurter/__init__.py
```diff
@@ -13,25 +13,26 @@
...
-
-    Returns the appropriate resource based on the provided parameters.
     """
-    # Determine which resource to return based on the `table` parameter
-    if table == "currencies":
-        return currencies()
-
-    elif table == "latest":
-        return latest()
+    dateTime = dlt.sources.incremental(
```
turtleDev, line 23: minor nitpick: python conventions prefer `snake_case` over `camelCase`. The only exception to this rule is `CapitalCase`, usually reserved for classes

ingestr/src/frankfurter/__init__.py
```diff
@@ -13,25 +13,26 @@
...
+        range_start="closed",
+        range_end="closed",
+    )
 
-    elif table == "exchange_rates":
-        return exchange_rates(start_date=start_date, end_date=end_date)
+    return (
+        currencies(),
+        latest(),
+        exchange_rates(start_date=dateTime, end_date=end_date),
```
turtleDev, line 34:
```suggestion
        exchange_rates(start_date=date_time, end_date=end_date),

```

turtleDev (approved):

PR #184, opened 2025-04-28
Google analytics: add real time report

sanjushahgupta:
User can fetch real time report based on dimensions, metrics and minutes range.
```
--source-uri: <realtime>:<dimensions>:<metrics>
--source-uri:<realtime>:<dimensions>:<metrics>:<minutes_range>
```
Screenshot 2025-04-29 at 10 02 24


ingestr/src/google_analytics/helpers.py
```diff
@@ -170,3 +239,54 @@ def _resolve_dimension_value(dimension_name: str, dimension_value: str) -> Any:
         return pendulum.from_format(dimension_value, "YYYYMMDDHHmm", tz="UTC")
     else:
         return dimension_value
+    
+def convert_minutes_ranges_to_int_list(minutes_ranges: str) -> List[int]:
```
karakanb, line 243: I wouldn't return this as a list of integers, instead I would make this return already a list of minute-range objects. that way if google supports more minute ranges tomorrow this would work automatically.

ingestr/src/sources.py
```diff
@@ -1455,48 +1455,45 @@ def handles_incrementality(self) -> bool:
...
-        parse_uri = urlparse(uri)
-        source_fields = parse_qs(parse_uri.query)
-        cred_path = source_fields.get("credentials_path")
-        cred_base64 = source_fields.get("credentials_base64")
-
-        if not cred_path and not cred_base64:
+        import ingestr.src.google_analytics.helpers as helpers
+        result = helpers.parse_google_analytics_uri(uri)
+        credentials = result["credentials"]
+        property_id = result["property_id"][0]
```
karakanb, line 1461: if you already have a helper function then that should return a proper response, which means if you'll anyway take just the first property id then just return one instead of a list.

ingestr/src/google_analytics/helpers.py
```diff
@@ -170,3 +222,72 @@ def _resolve_dimension_value(dimension_name: str, dimension_value: str) -> Any:
...
+
+    if "-" not in minutes_ranges:
+        raise ValueError(
+            "Invalid input. Minutes range should be startminute-endminute format. For example: 1-2,5-6"
+        )
+
+    if minutes_ranges.count("-") > 2 or minutes_ranges.count(",") > 1:
+        raise ValueError(
+            "You can define up to two time minutes ranges, formatted as comma-separated values `0-5,25-29`"
+        )
```
karakanb, line 244: no need for this, let them get the error from the API instead.

ingestr/src/google_analytics/helpers.py
```diff
@@ -170,3 +222,72 @@ def _resolve_dimension_value(dimension_name: str, dimension_value: str) -> Any:
...
+
+    minute_ranges = []
+    for min in minutes:
+        parts = min.split("-")
+        if len(parts) != 2 or not parts[0] or not parts[1]:
+            raise ValueError(
+                "Invalid input. Minutes range should be startminute-endminute format. For example: 1-2,5-6"
+            )
+        for part in parts:
+            minute_ranges.append(int(part))
```
karakanb, line 254: you don't need the second loop below, add them to the minute_ranges directly as objects here.

karakanb (approved):

PR #188, opened 2025-05-06
Add PhantomBuster as a Source

sanjushahgupta:


ingestr/src/phantombuster/client.py
```diff
@@ -0,0 +1,57 @@
...
+                "mode": "all",
+            }
+
+            if before_ended_at:
+                params["beforeEndedAt"] = before_ended_at
+
+            response = session.get(url=url, headers=self._get_headers(), params=params)
+            data = response.json()
+            containers = data.get("containers", [])
+            ended_times = []
```
sabrikaragonen, line 33: instead of an array, you can make it a variable

ingestr/src/phantombuster/client.py
```diff
@@ -30,23 +31,29 @@ def fetch_containers_result(self, session: requests.Session, agent_id: str):
             response = session.get(url=url, headers=self._get_headers(), params=params)
             data = response.json()
             containers = data.get("containers", [])
-            
+            started_at = start_date.int_timestamp * 1000 + int(start_date.microsecond / 1000)
```
sabrikaragonen, line 34: do you need to do it in while? you can take it above while loop I guess.

ingestr/src/phantombuster/client.py
```diff
@@ -13,7 +14,7 @@ def _get_headers(self):
             "accept": "application/json",
         }
 
-    def fetch_containers_result(self, session: requests.Session, agent_id: str):
+    def fetch_containers_result(self, session: requests.Session, agent_id: str, start_date: pendulum.DateTime, end_date: pendulum.DateTime):
         url = "https://api.phantombuster.com/api/v2/containers/fetch-all/"
         before_ended_at = None
         limit = 1000
```
sabrikaragonen, line 20: you can make limit 100

ingestr/src/phantombuster/client.py
```diff
@@ -30,23 +31,29 @@ def fetch_containers_result(self, session: requests.Session, agent_id: str):
             response = session.get(url=url, headers=self._get_headers(), params=params)
             data = response.json()
             containers = data.get("containers", [])
-            
+            started_at = start_date.int_timestamp * 1000 + int(start_date.microsecond / 1000)
+            ended_at = end_date.int_timestamp * 1000 + int(end_date.microsecond / 1000)
             for container in containers:
```
sabrikaragonen, line 36: if container status is running, skip it

ingestr/src/phantombuster/client.py
```diff
@@ -30,23 +31,29 @@ def fetch_containers_result(self, session: requests.Session, agent_id: str):
...
+                container_ended_at = container.get("endedAt")
+                if container_ended_at is None:
```
sabrikaragonen, line 38: if ended at is null, you can use start date (for failed ones, there was no endedAt I guess, so you can use startedAt)

ingestr/src/phantombuster/client.py
```diff
@@ -30,23 +31,29 @@ def fetch_containers_result(self, session: requests.Session, agent_id: str):
...
+                    continue
+                if not (started_at <= container_ended_at <= ended_at):
+                    continue
                 try:
                     result = self.fetch_result_object(session, container["id"])
-                    row = {"container": container, "result": result}
+                    partition_dt = pendulum.from_timestamp(container_ended_at / 1000, tz="UTC").to_date_string()
```
sabrikaragonen, line 44: same for here, if there is no endedAt, you can use startedAt

ingestr/src/phantombuster/client.py
sanjushahgupta: According to our discussion, we don't need this.

ingestr/src/phantombuster/client.py
sanjushahgupta: According to our discussion, we don't need this.

ingestr/src/phantombuster/client.py
sanjushahgupta: According to our discussion, we don't need this.

sabrikaragonen (approved):

PR #192, opened 2025-05-09
Added base currency functionality to Frankfurter

jbrand-dsp:
Users can now specify the base currency from Frankfurter's currency list for Frankfurter by adding "?={base_currency}" to the URI. 

The default base currency is USD if the base currency is not specified. 

The input is checked for compliance with the list of available currencies in Frankfurter.

The docs have been updated to reflect this functionality. Also currency_name was corrected to currency_code where necessary in the docs.


ingestr/src/sources.py
```diff
@@ -2179,7 +2179,22 @@ def dlt_source(self, uri: str, table: str, **kwargs):
             raise ValueError(
                 "Frankfurter takes care of incrementality on its own, you should not provide incremental_key"
             )
+
```
jbrand-dsp, line 2182: Minor reordering of imports so that all imports are together in one section

ingestr/src/frankfurter/__init__.py
```diff
@@ -92,21 +97,24 @@ def latest() -> Iterator[dict]:
             "date": date,
             "currency_code": currency_code,
             "rate": rate,
+            "base_currency": base_currency,
         }
 
 
 @dlt.resource(
-    write_disposition="merge",
+    write_disposition="replace",
```
turtleDev, line 105: any reason to use the `replace` disposition? this will replace the data in the destination on each run

ingestr/src/frankfurter/helpers.py
```diff
@@ -6,6 +6,40 @@
 
 FRANKFURTER_API_URL = "https://api.frankfurter.dev/v1/"
 
+CURRENCY_LIST = (
```
turtleDev, line 9: This method for validation can be fragile, if frankfurter updates the list of currencies they accept. You're better of hitting the API to validate whether a currency exists or not

jbrand-dsp: Good point! Didn't think of that!

ingestr/src/frankfurter/__init__.py
jbrand-dsp: Ups! A remnant of some testing...

turtleDev (approved):

PR #199, opened 2025-05-20
Add Attio as a Source

sanjushahgupta:


ingestr/src/attio/helpers.py
```diff
@@ -0,0 +1,80 @@
...
+        self, url: str, client: requests.Session, limit: int = 1000, params=None
+    ):
+        if params is None:
+            params = {}
+        offset = 0
+        while True:
+            query_params = {**params, "limit": limit, "offset": offset}
+            response = client.get(url, headers=self.headers, params=query_params)
+
+            data = response.json()["data"]
```
sabrikaragonen, line 25: what if there is no data key? it would fail. I'd add a check here first to avoid an exception

ingestr/src/attio/helpers.py
```diff
@@ -0,0 +1,80 @@
...
+    item["workspace_id"] = item["id"]["workspace_id"]
+    if item["id"].get("object_id") is not None:
+        item["object_id"] = item["id"]["object_id"]
+    if item["id"].get("record_id") is not None:
+        item["record_id"] = item["id"]["record_id"]
+    if item["id"].get("list_id") is not None:
+        item["list_id"] = item["id"]["list_id"]
+    if item["id"].get("entry_id") is not None:
+        item["entry_id"] = item["id"]["entry_id"]
+    item["partition_dt"] = pendulum.parse(item["created_at"]).date()  # type: ignore
```
sabrikaragonen, line 79: it's a bit magic, we should know what to expect for each endpoint and define them seperately. 
Also, we don't need partition_dt column. We  can just use created_at itself.

ingestr/src/sources.py
```diff
@@ -2392,3 +2392,56 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+                    "Lists table must be in the format `list_entries:list_id`"
+                )
+            list_id = table_name[1]
+        elif table_name[0] == "all_list_entries":
+            if len(table_name) != 2:
+                raise ValueError(
+                    "All list entries table must be in the format `all_list_entries:object_api_slug`"
+                )
+            object_id = table_name[1]
+
```
sabrikaragonen, line 2440: for each resource, we have logic in 3 different places.
can't we push this part to the resource?

sabrikaragonen (changes requested):

ingestr/src/attio/helpers.py
```diff
@@ -11,39 +12,27 @@ def __init__(self, api_key: str):
...
+                )
+            else:
+                response = self.client.post(
+                    url, headers=self.headers, params=query_params
+                )
 
-            if len(data) < limit:
+            response_data = response.json()
+            if "data" not in response_data:
                 break
```
sabrikaragonen, line 34: not a break, but error maybe? it's not the expected output, so we should warn the users somehow. we don't want to lose data

ingestr/src/attio/helpers.py
```diff
@@ -0,0 +1,67 @@
...
+
+            for item in data:
+                flat_item = flat_attributes(item)
+                yield flat_item
+
+            if len(data) < limit:
+                break
+            offset += limit
+
+    def fetch_all_list_entries_for_object(self, object_id: str | None):
```
sabrikaragonen, line 47: can we not move it to the resource. do we need a seperate function for that?

sabrikaragonen (approved):

ingestr/src/attio/helpers.py
```diff
@@ -0,0 +1,56 @@
+import pendulum
+
+from ..utility import create_client
```
turtleDev, line 3:
```suggestion
from from ingestr.src.utility import create_client
```

ingestr/src/utility.py
```diff
@@ -0,0 +1,18 @@
+import requests
```
turtleDev, line 1: using generic names like `utils` or `utility` is considered bad practice for packages. 

Rather, we should name it something more specific. `http_client` for instance is a much better name.

ingestr/src/utility.py
```diff
@@ -0,0 +1,18 @@
...
+from dlt.sources.helpers.requests import Client
+
+
+def create_client() -> requests.Session:
```
turtleDev, line 5: Love it that you've refactored a common piece of code to a standalone module.

It would be great if you could refactor other sources that duplicate this client. You may even want to parameterise this function so users of this package can pass in additional configuration to override the defaults.

ingestr/src/sources.py
```diff
@@ -2392,3 +2392,35 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+        api_key = query_params.get("api_key")
+
+        if api_key is None:
+            raise MissingValueError("api_key", "Attio")
+
+        parts = table.replace(" ", "").split(":")
+        table_name = parts[0]
+        params = parts[1:]
+
+        if table_name not in [
```
turtleDev, line 2413: There's a trick to validation resources that's more robust than this.

Once a source is constructed, you can reference it's `resources` field. This means that you never have to update this list, even if you add or remove resources in the future.

```py
src = attio_source(api_key=api_key[0], params=params)
if table_name not in src.resources:
  raise ...
```

turtleDev: here's an example: https://github.com/bruin-data/ingestr/blob/main/ingestr/src/sources.py#L1700

ingestr/src/attio/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+            for item in data:
+                flat_item = flatten_item(item)
+                yield flat_item
+
+            if len(data) < limit:
+                break
+            offset += limit
+
+
+def flatten_item(item: dict) -> dict:
```
karakanb, line 52: I think we should flatten `values` as well if they exist

ingestr/src/attio/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+            else:
+                response = self.client.post(
+                    url, headers=self.headers, params=query_params
+                )
+
+            if response.status_code != 200:
+                raise Exception(f"HTTP {response.status_code} error: {response.text}")
+
+            response_data = response.json()
+            if "data" not in response_data:
```
turtleDev, line 36: It might be a good idea to print the payload returned by the API to help in debugging the issue.

ingestr/src/utility.py
karakanb: if anything, I would move these functions into the other file, keep them simple

karakanb: I agree with the naming since this will be used across different sources

ingestr/src/utility.py
karakanb: agreed, although I suggest doing this in another PR

ingestr/src/attio/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+                flat_item = flatten_item(item)
+                yield flat_item
+
+            if len(data) < limit:
+                break
+            offset += limit
+
+
+def flatten_item(item: dict) -> dict:
+    if "id" in item:
```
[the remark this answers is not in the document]
sanjushahgupta: Sabri and I decided to keep id as it is.

ingestr/src/attio/helpers.py
sanjushahgupta: Sabri and I decided to flat id only

turtleDev (approved):

PR #215, opened 2025-06-01
Solidgate: add financial entries resource

sanjushahgupta:


ingestr/src/http_client.py
```diff
@@ -2,17 +2,21 @@
 from dlt.sources.helpers.requests import Client
 
 
-def create_client() -> requests.Session:
+def create_client(status_code: int = 502) -> requests.Session:
```
sabrikaragonen: status_code variable name is not clear. maybe we should use a clearerer name showing what it does

ingestr/src/solidgate/helpers.py
```diff
@@ -66,6 +69,67 @@ def fetch_data(
...
+        }
+
+        url = f"{self.base_url}/finance/financial_entries"
+        post_response = self.client.post(url, headers=headers, json=request_payload)
+        post_response.raise_for_status()
+        report_url = post_response.json().get("report_url")
+        if not report_url:
+            return f"Report URL not found in the response: {post_response.json()}", 400
+
+        time.sleep(5)
```
sabrikaragonen, line 96: why is there a sleep? maybe we should add a comment here to tell the reason

PR #228, opened 2025-06-07
Kinesis: Optionally address streams per StreamARN

amotl:
Dear @karakanb,

thanks a stack for starting ingestr. This patch brings in a minor improvement to address a Kinesis stream by ARN instead of name, like using `arn:aws:kinesis:eu-central-1:842404475894:stream/test` instead of just using `test`.

The patch was too lazy to introduce a dedicated parameter, so it modifies just a few bits to optionally accept a [StreamARN](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_StreamDescription.html#Streams-Type-StreamDescription-StreamARN) into the existing `stream_name` parameter without much ado. Do you think it's ok doing it like this? We think it offers excellent usability without needing to introduce yet another parameter.

With kind regards,
Andreas.

## References
- https://github.com/hampsterx/async-kinesis/pull/39

### DIFF
--- a/ingestr/src/kinesis/__init__.py
+++ b/ingestr/src/kinesis/__init__.py
@@ -42,7 +42,7 @@ def kinesis_stream(
         initial_at_timestamp (TAnyDateTime): An initial timestamp used to generate AT_TIMESTAMP or LATEST iterator when timestamp value is 0
         max_number_of_messages (int): Maximum number of messages to read in one run. Actual read may exceed that number by up to chunk_size. Defaults to None (no limit).
         milliseconds_behind_latest (int): The number of milliseconds behind the top of the shard to stop reading messages, defaults to 1000.
-        parse_json (bool): If True, assumes that messages are json strings, parses them and returns instead of `data` (otherwise). Defaults to False.
+        parse_json (bool): If True, assumes that messages are json strings, parses them and returns instead of `data` (otherwise). Defaults to True.
         chunk_size (int): The number of records to fetch at once. Defaults to 1000.
     Yields:
             Iterable[TDataItem]: Messages. Contain Kinesis envelope in `kinesis` and bytes data in `data` (if `parse_json` disabled)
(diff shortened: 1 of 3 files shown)


ingestr/src/kinesis/helpers.py
```diff
@@ -63,3 +63,20 @@ def max_sequence_by_shard(values: Sequence[StrStr]) -> StrStr:
...
+    - customer_events
+    - arn:aws:kinesis:eu-central-1:842404475894:stream/customer_events
+
+    https://docs.aws.amazon.com/kinesis/latest/APIReference/API_StreamDescription.html#Streams-Type-StreamDescription-StreamName
+    https://docs.aws.amazon.com/kinesis/latest/APIReference/API_StreamDescription.html#Streams-Type-StreamDescription-StreamARN
+    """
+    if stream_name.startswith("arn:"):
+        return {"StreamARN": stream_name}
+    else:
+        return {"StreamName": stream_name}
```
amotl, line 82: > Do you think it's ok doing it like this?

Well, it's the beauty of heuristics. Let me know if you don't like it, or if you want to see it done differently. By reusing the `stream_name` parameter, the patch is rather minimal, and we think it's also very convenient from a user's perspective.

amotl:
Hi @karakanb. Do you see a chance to merge this patch?

ingestr/src/kinesis/__init__.py
```diff
@@ -42,7 +42,7 @@ def kinesis_stream(
         initial_at_timestamp (TAnyDateTime): An initial timestamp used to generate AT_TIMESTAMP or LATEST iterator when timestamp value is 0
         max_number_of_messages (int): Maximum number of messages to read in one run. Actual read may exceed that number by up to chunk_size. Defaults to None (no limit).
         milliseconds_behind_latest (int): The number of milliseconds behind the top of the shard to stop reading messages, defaults to 1000.
-        parse_json (bool): If True, assumes that messages are json strings, parses them and returns instead of `data` (otherwise). Defaults to False.
+        parse_json (bool): If True, assumes that messages are json strings, parses them and returns instead of `data` (otherwise). Defaults to True.
```
amotl, line 45: FYI: Unrelated to the patch, this is just fixing a possible leftover.

karakanb (approved):

PR #231, opened 2025-06-07
Dependencies: Relax version pinning on `requests`

amotl:
## Problem
We found ingestr does not cooperate well with other packages using `requests`.

## Solution
Relax its dependency version pinning a bit.


amotl:
Closing this in favor of a more generic approach.

- GH-239

PR #232, opened 2025-06-07
CrateDB: Add database source adapter

amotl:
## About

The patch unlocks [CrateDB](https://github.com/crate/crate) as a database source. Thanks!

## Preview

Using this command, you can run the [corresponding example](https://github.com/bruin-data/ingestr/blob/0a0b4a31db194b8c43e772a9ea259710813d6b47/docs/supported-sources/cratedb.md#example) without much ado.
```shell
alias ingestr='uvx --with="ingestr @ git+https://github.com/crate-workbench/ingestr@cratedb-source" ingestr'
```

## Details

CrateDB's source adapter is based on SQLAlchemy, using its [sqlalchemy-cratedb](https://pypi.org/project/sqlalchemy-cratedb/) dialect package.

## References
- GH-238
- https://github.com/crate/crate-clients-tools/issues/86


docs/supported-sources/cratedb.md
```diff
@@ -0,0 +1,69 @@
...
+    --dest-table 'dest.summits'
+```
+```shell
+duckdb cratedb.duckdb 'SELECT * FROM dest.summits LIMIT 5'
+```
+
+<img alt="CrateDB_img" src="../media/cratedb-source.png" />
+
+> [!WARNING]
+> CrateDB currently doesn't support `delete+insert`, `merge` or `scd2` incremental strategies.
```
amotl: @hlcianfagna: I've blatantly copied this from the Clickhouse documentation, without knowing too much about corresponding details. As I think you have expertise in this area, could you please validate this statement in the context of ingestr and CrateDB, possibly including some hands-on validations? 🙏

docs/supported-sources/cratedb.md
```diff
@@ -0,0 +1,69 @@
...
+
+<!--
+    see https://github.com/crate/crate-clients-tools/issues/86#issuecomment-2466516366
+-->
```
amotl: This tracks a few incompatibility issues. I don't know how this will be rendered on the documentation. Maybe let's slot it into a visible paragraph into an "appendix" section on this page, or remove altogether.

- https://github.com/crate/crate-clients-tools/issues/86#issuecomment-2466516366

docs/supported-sources/cratedb.md
amotl: At least I can tell from my basic orientation flights that I have been able to use `Incremental Strategy: replace` successfully.

image

This is probably not a big surprise, as it doesn't need any database support at all?

amotl: Doesn't this detail apply to the destination adapter rather than the source adapter? If you agree, it should be made a concern of GH-237 instead.

hlcianfagna: I was just just about to write that, I believe that only makes a difference for destination, not sources.

amotl:
@karakanb: Thanks for your kind response. Yes, CrateDB offers a Docker container, and there is also an [adapter](https://github.com/crate/cratedb-toolkit/blob/main/cratedb_toolkit/testing/testcontainers/cratedb.py) for [Testcontainers Python](https://pypi.org/project/testcontainers/), yet pending a standalone release.

We haven't been able to discover integration tests within the code base of ingestr. If you can point out how we could contribute any, we will be happy to do so.

docs/supported-sources/cratedb.md
amotl: Thanks. I will carry this detail over to GH-237 to discuss it there later.

amotl: I've moved the discussion to the other PR.

- https://github.com/bruin-data/ingestr/pull/237#discussion_r2135911178

docs/supported-sources/cratedb.md
```diff
@@ -0,0 +1,80 @@
...
+```shell
+docker run --rm -it --name=cratedb \
+  --publish=4200:4200 --publish=5432:5432 \
+  --env=CRATE_HEAP_SIZE=2g crate/crate:nightly \
+  -Cdiscovery.type=single-node
+```
+
+We are tracking development progress and incompatibilities at 
+[Support for ingestr/CrateDB] and ["tool: dlt/ingestr"]. Please join the discussion
+or share relevant issue reports that help us improve interoperability. Thanks!
```
amotl, line 73: The appendix now includes this auxiliary information, including a few tidbits you had asked about.

- NB: See also https://github.com/bruin-data/ingestr/issues/238#issuecomment-2956589872.

docs/supported-sources/cratedb.md
amotl: Added this to an "appendix" section, as suggested.

- https://github.com/bruin-data/ingestr/pull/232#pullrequestreview-2910929280

karakanb (approved):

PR #277, opened 2025-07-03
Dependencies: Downgrade to `sqlalchemy-cratedb==0.41.0` GA

amotl:
## Problem

When using `uv pip install --upgrade ingestr`, only version v0.13.52 gets installed.

## Details

Adding support for CrateDB as a source adapter per GH-232 added the dependency `sqlalchemy-cratedb==0.42.0.dev2`, which is apparently problematic in some cases.

- https://github.com/crate-workbench/ingestr/issues/1#issuecomment-3033703828

To install the latest version of ingestr with `sqlalchemy-cratedb==0.42.0.dev2` successfully, `uv` needs the `--prerelease=allow` option. Otherwise, it will install the much older ingestr version v0.13.52.

## Solution 

Downgrade to a previous GA version of sqlalchemy-cratedb (v0.41.0), to mitigate any such obstacles.

## Outlook

Note to self: We'll need to wrap up another patch'es ingredients to fully conclude the operation making 0.42.0 available.

- https://github.com/crate/sqlalchemy-cratedb/pull/215

## References
- GH-292

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -25,7 +25,7 @@ rich==13.9.4
 rudder-sdk-python==2.1.4
 snowflake-sqlalchemy==1.6.1
 sqlalchemy-bigquery==1.12.1
-sqlalchemy-cratedb==0.42.0.dev2
+sqlalchemy-cratedb==0.41.0
 sqlalchemy-hana==2.0.0
 sqlalchemy-redshift==0.8.14
 SQLAlchemy==1.4.52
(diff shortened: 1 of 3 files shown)


amotl:
> To install the latest version of ingestr with `sqlalchemy-cratedb==0.42.0.dev2` successfully, `uv` needs the `--prerelease=allow` option. Otherwise, it will install the much older ingestr version v0.13.52.

In this spirit, the impact is also on users not using the CrateDB adapter. Apologies for introducing this hiccup previously.

turtleDev (approved):

amotl:
Thanks a stack for merging and releasing so swiftly! 💯

PR #281, opened 2025-07-04
mssql:  add support for token auth

turtleDev:
This change extends SQL Server source and destination to accept access token generated by `az` cli.

We re-purpose the deprecated `ActiveDirectoryAccessToken` Authentication parameter as an indication that the password is a token.

example:
```bash
TOKEN=$(az account get-access-token --resource https://database.windows.net/ --query accessToken -o tsv)
USER=$(az account show --query user.name -o tsv)
SERVER="sql-server-expo"
DB="ingestr"
CONN_STR="mssql://$USER:$TOKEN@$SERVER.database.windows.net/$DB?driver=ODBC+Driver+18+for+SQL+Server&authentication=ActiveDirectoryAccessToken&encrypt=yes"

python3 -m ingestr.main ingest \
    --source-uri "$CONN_STR" \
    --source-table "dbo.example" \
    --dest-uri "$CONN_STR" \
    --dest-table "dbo.example_3" \
    --yes
```

### DIFF
--- a/ingestr/src/destinations.py
+++ b/ingestr/src/destinations.py
@@ -5,6 +5,8 @@ import json
 import os
 import shutil
 import tempfile
+import struct
+import datetime
 from urllib.parse import parse_qs, quote, urlparse
 
 import dlt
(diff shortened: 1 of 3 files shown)


ingestr/src/destinations.py
```diff
@@ -143,9 +155,89 @@ def dlt_dest(self, uri: str, **kwargs):
...
+class MsSqlDestImpl(dlt.destinations.mssql):
+    @property
+    def client_class(self):
+        return MsSqlClient
+
+
 class MsSQLDestination(GenericSqlDestination):
     def dlt_dest(self, uri: str, **kwargs):
-        return dlt.destinations.mssql(credentials=uri, **kwargs)
+        return MsSqlDestImpl(credentials=uri, **kwargs)
```
amotl, line 240: Hi. We found this spot breaks dependency isolation wrt. pyodbc.

- https://github.com/bruin-data/ingestr/issues/293#issuecomment-3044922655

PR #284, opened 2025-07-05
Tests: Adjust software tests for CrateDB

amotl:
## About
Retroactively add software tests for CrateDB by adjusting the existing test suite.

_It is very unlikely the patch will get merged in its current form, so it will need to be maintained separately to provide quality assurance for the CrateDB source and destination adapters._

## Thoughts
The patch clearly demonstrates where CrateDB is a different animal compared to other DBMS. We don't know yet how to compress or re-shape those anomaly-balancing boilerplate adjustments, in order to reduce the noise of the patch, but we hope to be able to resolve one or another obstacle for the better as we go.

## Backlog
- [x] After making a start per `main_test.py`, expand into ~`sources_test.py` and~ `destinations_test.py` as well.
- [ ] Get rid of the worst code smells, by fixing or improving the dlt adapter, or by providing relevant monkey patches.
- [ ] Release CrateDB Testcontainer for Python to get rid of development dependency to `cratedb-toolkit`.

## References
- GH-238

PR #287, opened 2025-07-06
Tests: Improve isolation through resource teardown by closing/disposing connections etc.

amotl:
## About

This patch has been broken out of GH-284 in the same spirit like GH-283, supporting to run combinations of individual test cases without using `xdist`.

Our evaluations are that it feels like software tests have not been explicitly isolated yet, so `xdist` came to the rescue for you, because it implicitly isolates test cases by distributing them amongst worker processes?

## Background

Most prominently, adjustments in the area of DuckDB connection handling get rid of this error.

**`Connection Error: Can't open a connection to same database file with a different configuration than existing connections`**

```python
dlt.destinations.exceptions.DestinationConnectionError: Connection with DuckDbSqlClient to dataset name 
foo failed. Please check if you configured the credentials at all and provided the right credentials values.
You can be also denied access or your internet connection may be down. The actual reason given is:
Connection Error: Can't open a connection to same database file with a different configuration than existing connections
```
-- https://github.com/bruin-data/ingestr/actions/runs/16102670949/job/45433825506#step:11:607

## Notes

> [...] without using `xdist`.

On the workbench, we are running individual software tests like this to speed up cycle time by avoiding `xdist` startup overhead, that's why GH-283 came up.

```shell
export TESTCONTAINERS_RYUK_DISABLED=true

time pytest -vvv -k "(replace or append) and (duck and postgres)"
time pytest -vvv -k "(duck and postgres)"

time pytest -vvv -k "(dynamo and postgres) or (gcs and clickhouse)"
time pytest -vvv -k "cratedb and not sqlserver"
```

### DIFF
--- a/ingestr/main_test.py
+++ b/ingestr/main_test.py
@@ -245,6 +245,8 @@ def test_create_replace_csv_to_duckdb():
     for i, row in enumerate(actual_rows):
         assert res[i] == tuple(row)
 
+    # Clean up
+    conn.close()
     try:
         os.remove(abs_db_path)
     except Exception:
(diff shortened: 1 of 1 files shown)


amotl:
This PR also needs to include 3deb686f82d8.

PR #289, opened 2025-07-06
Naming things: Rename `_kafka_msg_id` to `_kafka__msg_id`

amotl:
## About
Unfortunately, CrateDB does not permit column names using leading underscores, like `_kafka_msg_id`. However, if two underscores are included, like `_kafka__msg_id`, it works.

The patch improves storage alignment with the other already existing column `_kafka__data`, which is already using two underscores in its name.

## Thoughts
We certainly know the obstacle is on CrateDB, while it is not a problem for any other database destination covered by ingestr.

However, we didn't see many dependencies on this column name, so we are humbly asking if you could do us a favor to rename that column, so we will not need to find a different workaround in form of a runtime patch. 🙏 

## References
- https://github.com/crate-workbench/ingestr/issues/3

### DIFF
--- a/ingestr/main_test.py
+++ b/ingestr/main_test.py
@@ -1293,7 +1293,7 @@ def test_kafka_to_db(dest):
         dest_engine = sqlalchemy.create_engine(dest_uri)
         with dest_engine.connect() as conn:
             res = conn.execute(
-                "select _kafka__data from testschema.output order by _kafka_msg_id asc"
+                "select _kafka__data from testschema.output order by _kafka__msg_id asc"
             ).fetchall()
         dest_engine.dispose()
         return res
--- a/ingestr/src/kafka/helpers.py
+++ b/ingestr/src/kafka/helpers.py
@@ -46,7 +46,7 @@ def default_msg_processor(msg: Message) -> Dict[str, Any]:
             },
             "data": msg.value().decode("utf-8"),
         },
-        "_kafka_msg_id": digest128(topic + str(partition) + str(key)),
+        "_kafka__msg_id": digest128(topic + str(partition) + str(key)),
     }
 
 


amotl:
We think the new field `_kafka__msg_id` is much better aligned with the others now.

```sql
CREATE TABLE IF NOT EXISTS "doc"."kafka_demo" (
   "_kafka__partition" BIGINT,
   "_kafka__topic" TEXT,
   "_kafka__offset" BIGINT,
   "_kafka__ts__type" BIGINT,
   "_kafka__ts__value" TIMESTAMP WITH TIME ZONE,
   "_kafka__data" TEXT,
   "_kafka__msg_id" TEXT,
   "__dlt_load_id" TEXT NOT NULL,
   "__dlt_id" TEXT NOT NULL
);
```

amotl:
Hi Burak.

> This seems like a breaking change for people that are using ingestr with Kafka.

That's true. Let's do it more conservatively to not introduce breaking changes without previously emitting warnings. So, the primary question is: Do you agree with the change in general, in order to align the fields on matters of "naming things"? If so, let's publish this change progressively, like any good software is doing it? I will be happy to bring in corresponding changes to this patch. If you don't like it at all, let's just close the PR.

> Could this be implemented as an adapter that modifies the data that is only registered for CrateDB?

Yes, as it's software, many things are possible. I often use monkeypatching for such matters, only changing code paths in certain situations, optimally by using a context manager so the change doesn't pollute the runtime context permanently, thanks to Python.

In this case, because CrateDB currently uses a custom dlt destination adapter [dlt-cratedb](https://github.com/crate/dlt-cratedb), we already added a corresponding adjustment per https://github.com/crate/dlt-cratedb/commit/7c146329f94ebf387f416fb33fd9a15b909200b3. In this spirit, we absolutely do not have an immediate need for this change at all, so the case here is effectively meant as a long-term improvement, which leads to my previous question:

> If you agree with the change in general, in order to align the fields on matters of "naming things", let's publish this change progressively, by first bringing in a relevant `UserWarning`?

With kind regards,
Andreas.

amotl:
Another option would be to use a dedicated URL parameter on the Kafka source, when applicable? Silly idea?

amotl:
We submitted a patch that can satisfy new requirements (together with other features) while retaining backwards compatibility.

- GH-306

amotl:
Closing this in favor of GH-289.

Issue #293, opened 2025-07-07
Problem with v0.13.64 and v0.13.65: `ModuleNotFoundError: No module named 'pyodbc'`


amotl:
## Problem

When running ingestr using `uvx --prerelease=allow 'ingestr>=0.13.64' ingest ...`, it fails.
```python
ModuleNotFoundError: No module named 'pyodbc'
```

## Interim solution

⚠ Workaround ⚠

Please downgrade to a previous version of ingestr, e.g. by skipping the offending versions v0.13.64 and v0.13.65,
```
uv pip install --prerelease=allow --upgrade 'ingestr!=0.13.64,!=0.13.65'
```
or install ingestr including its `odbc` extra.
```
uv pip install --prerelease=allow --upgrade 'ingestr[odbc]'
```

## Permanent solution

Please [yank](https://docs.pypi.org/project-management/yanking/) the [ingestr v0.13.64](https://pypi.org/project/ingestr/0.13.64/) and [ingestr v0.13.65](https://pypi.org/project/ingestr/0.13.65/) packages from PyPI.

## Details

```python

  File "/Users/amo/.cache/uv/archive-v0/XoWBb3yE3D5q7XjrnFwMr/lib/python3.13/site-packages/ingestr/main.py", line 291, in ingest
    from ingestr.src.destinations import AthenaDestination
  File "/Users/amo/.cache/uv/archive-v0/XoWBb3yE3D5q7XjrnFwMr/lib/python3.13/site-packages/ingestr/src/destinations.py", line 22, in <module>
    from dlt.destinations.impl.mssql.mssql import (
    ...<2 lines>...
    )
  File "/Users/amo/.cache/uv/archive-v0/XoWBb3yE3D5q7XjrnFwMr/lib/python3.13/site-packages/dlt/destinations/impl/mssql/mssql.py", line 15, in <module>
    from dlt.destinations.impl.mssql.sql_client import PyOdbcMsSqlClient
  File "/Users/amo/.cache/uv/archive-v0/XoWBb3yE3D5q7XjrnFwMr/lib/python3.13/site-packages/dlt/destinations/impl/mssql/sql_client.py", line 6, in <module>
    import pyodbc
ModuleNotFoundError: No module named 'pyodbc'
```

amotl:
When adding the `odbc` extra and going back to Python 3.12, using `uvx --python=3.12 --prerelease=allow 'ingestr[odbc]>=0.13.64' ingest ...`, it works.

[...]

PR #296, opened 2025-07-07
Dependencies: Isolate `pyodbc` package

amotl:
## About
The most recent release v0.13.64 made the [pyodbc](https://pypi.org/project/pyodbc/) package a strict requirement. This patch relaxes that again by isolating the dependency into a dedicated module which is only imported at runtime when MSSQL is selected.

## Synopsis
The application will not fail when the pyodbc package is not installed.
```shell
(.venv) sink:ingestr amo$ uv pip uninstall pyodbc
Uninstalled 1 package in 44ms
 - pyodbc==5.2.0
(.venv) sink:ingestr amo$ python
Python 3.13.5 (main, Jun 11 2025, 15:36:57) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ingestr.src.destinations
```

## Details
The isolation is accompanied by a minor testing/CI reorg to cover both situations well. Introducing a dedicated `full` extra which bundles all the extra dependency groups into a single entrypoint has been approved in other polyglot code bases in the past.

It is well suited to cover two scenarios to verify if the package works with all extras, and without any extras. Of course, this paradigm can be made more granular when there is demand. As a common best practice, discriminating between minimal vs. full is good enough.

## References
- GH-293

### DIFF
--- a/ingestr/src/destinations.py
+++ b/ingestr/src/destinations.py
@@ -1,30 +1,18 @@
 import abc
 import base64
 import csv
-import datetime
 import json
 import os
 import shutil
-import struct
 import tempfile
 from urllib.parse import parse_qs, quote, urlparse
 
 import dlt
 import dlt.destinations.impl.filesystem.filesystem
 from dlt.common.configuration.specs import AwsCredentials
-from dlt.common.destination.capabilities import DestinationCapabilitiesContext
-from dlt.common.schema import Schema
 from dlt.common.storages.configuration import FileSystemCredentials
 from dlt.destinations.impl.clickhouse.configuration import (
     ClickHouseCredentials,
-)
-from dlt.destinations.impl.mssql.configuration import MsSqlClientConfiguration
-from dlt.destinations.impl.mssql.mssql import (
-    HINT_TO_MSSQL_ATTR,
-    MsSqlJobClient,
-)
-from dlt.destinations.impl.mssql.sql_client import (
-    PyOdbcMsSqlClient,
 )
 
 from ingestr.src.errors import MissingValueError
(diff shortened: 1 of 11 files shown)


ingestr/src/destinations.py
```diff
@@ -155,88 +143,12 @@ def dlt_dest(self, uri: str, **kwargs):
...
-        return MsSqlClient
-
-
 class MsSQLDestination(GenericSqlDestination):
     def dlt_dest(self, uri: str, **kwargs):
+        from ingestr.src.destinations_mssql import (  # type: ignore[import-untyped]
+            MsSqlDestImpl,
+        )
+
         return MsSqlDestImpl(credentials=uri, **kwargs)
```
amotl: I don't know why this type hint override is needed: The module file is physically present within the same package, and we only received such an error message from mypy when pulling in modules from 3rd party packages. That's why we don't know what to make out of it, so we slapped a type-ignore on the line.

```
(.venv) sink:ingestr amo$ make lint-ci
```
```
ingestr/src/destinations.py:148: error: Skipping analyzing "ingestr.src.destinations_mssql": module is installed, but missing library stubs or py.typed marker  [import-untyped]
ingestr/src/destinations.py:148: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 74 source files)
make: *** [Makefile:36: lint-ci] Error 1
```

ingestr/src/sources.py
```diff
@@ -267,7 +267,7 @@ def table_rows(
                 import pyodbc  # type: ignore
                 from sqlalchemy import create_engine
 
-                from ingestr.src.destinations import (
+                from ingestr.src.destinations_mssql import (  # type: ignore[import-untyped]
```
amotl: Dito.

pyproject.toml
```diff
@@ -148,11 +148,14 @@ classifiers = [
...
+full = [
+    "ingestr[oracle,odbc]",
+]
 oracle = [
     "cx_Oracle==8.3.0",
 ]
 odbc = [
-    "pyodbc==5.1.0",
+    "pyodbc==5.2.0",
 ]
```
amotl, line 159: `requirements-dev.txt` said 5.2.0 would be fine, so we used it without much ado.

turtleDev:
fixed by https://github.com/bruin-data/ingestr/pull/298

amotl:
Hi @turtleDev. Feel free to close PR if you don't have a need.

My advise is to revert GH-298 and use this patch instead, because it also takes care about relevant software testing matters, and also takes a clean approach instead of mixing concerns.

- https://github.com/bruin-data/ingestr/pull/298#issuecomment-3046081292

requirements-dev.txt
```diff
@@ -6,7 +6,6 @@ pytest==8.3.3
 ruff==0.11.4
 hatchling==1.27.0
 build==1.2.1
-pyodbc==5.2.0
 twine==6.0.1
 testcontainers[postgres,mysql]==4.8.2
 pytest-xdist[psutil]==3.6.1
```
amotl, line 11: Removing `pyodbc` here is absolutely crucial so that its presence will not negatively mask any operations on the code base to make it work without.

turtleDev:
hey @amotl it's not a (high) priority right now for us. There are some rough edges around ingestr that we plan to refactor down the road. Thanks for the contribution though!

I would like to invite you to join our [slack community](https://join.slack.com/t/bruindatacommunity/shared_invite/zt-2dl2i8foy-bVsuMUauHeN9M2laVm3ZVg). You can get a better picture of what we're working on and as an alternate platform for discussions around ingestr.

amotl:
Hi @turtleDev. Then let's keep the PR open for the future refactorings you are referring to, that could lead to corresponding quality improvements?

Thanks for the invite, I may join your Slack, but in general I try to keep distraction low and productivity high, and there is already another Slack I need to deal with ;].

Let's put the PR back into draft mode?

PR #300, opened 2025-07-09
add influx db as a source

sanjushahgupta:


ingestr/main.py
```diff
@@ -596,8 +597,14 @@ def set_primary_key(x):
         if incremental_strategy != IncrementalStrategy.none:
             write_disposition = incremental_strategy.value
 
-        start_time = datetime.now()
+        if factory.source_scheme == "influxdb":
```
turtleDev, line 600: if I remember correctly, merge falls back to append automatically if a primary key is not specified.

sanjushahgupta: Yes, it will append if a primary key is not present.

PR #301, opened 2025-07-09
feat: make snowflake sources compatiable with key based auth

terzioglub:


ingestr/src/sources.py
```diff
@@ -140,6 +161,10 @@ def eng_callback(engine):
 
             engine_adapter_callback = eng_callback
 
+        # if uri.startswith("snowflake://"):
+        #     from dlt.sources.sql_database.helpers import BigQueryAdapter
+        #
```
karakanb: remove these please

PR #304, opened 2025-07-10
OCI: Slightly streamline the `Dockerfile`

amotl:
That's just a few bits of copy-editing the `Dockerfile`.

- Use canonical OCI-based installation method for `uv`
- Use environment variables to configure `uv`
- Remove redundant platform-switch instructions (x64 vs. arm64)
- Compress `uv pip install` into a single incantation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,27 @@
 FROM python:3.11-slim
 
+# Guidelines that have been followed.
+# - https://hynek.me/articles/docker-uv/
+
+# Install the `uv` package manager.
+# Security-conscious organizations should package/review uv themselves.
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
+
+# - Tell uv to byte-compile packages for faster application startups.
+# - Silence uv complaining about not being able to use hard links.
+# - Prevent uv from accidentally downloading isolated Python builds.
+# - Install packages into the system Python environment.
+ENV \
+    UV_COMPILE_BYTECODE=1 \
+    UV_LINK_MODE=copy \
+    UV_PYTHON_DOWNLOADS=never \
+    UV_SYSTEM_PYTHON=1
+
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,14 +1,27 @@
...
+
+# - Tell uv to byte-compile packages for faster application startups.
+# - Silence uv complaining about not being able to use hard links.
+# - Prevent uv from accidentally downloading isolated Python builds.
+# - Install packages into the system Python environment.
+ENV \
+    UV_COMPILE_BYTECODE=1 \
+    UV_LINK_MODE=copy \
+    UV_PYTHON_DOWNLOADS=never \
+    UV_SYSTEM_PYTHON=1
```
amotl, line 18: Using `UV_COMPILE_BYTECODE=1` is the main attraction of this patch. ;]

Dockerfile
```diff
@@ -1,14 +1,27 @@
...
+
 WORKDIR /app
 
-COPY ./requirements.txt /app/requirements.txt
-COPY ./requirements_arm64.txt /app/requirements_arm64.txt
-RUN if [ "$(uname -m)" = "aarch64" ]; then \
-    cp /app/requirements_arm64.txt /app/requirements.txt; \
-fi
```
karakanb, line 9: why were these removed?

Dockerfile
```diff
@@ -28,17 +41,18 @@ RUN \
...
-RUN /install.sh && rm /install.sh
 
-RUN $HOME/.local/bin/uv pip install --system --no-cache -r requirements.txt
+# Install application.
 
+# Copy sources and activate platform-specific requirements file.
 COPY . /app
 RUN if [ "$(uname -m)" = "aarch64" ]; then \
     cp /app/requirements_arm64.txt /app/requirements.txt; \
 fi
```
amotl, line 51: Hi. The lines above have been redundant. These here are valid, and are doing their job perfectly well. Now, a little inline comment was added.

PR #306, opened 2025-07-13
Improve Kafka adapter to provide features closer to Kafka Connect

amotl:
## Pitch

Expand Kafka adapter's `default_msg_processor` into a miniature decoding unit.

Loading data from Kafka into a database destination works well, but we found there are no options to specifically _decode and drill down into_ the **Kafka event value** properly, in order to only relay that into the target database, without any metadata information.

## Comparison

For example, [Kafka Connect](https://www.confluent.io/blog/kafka-connect-deep-dive-converters-serialization-explained/) provides such configuration options for similar use cases which are fragments hereof.
```json
"key.converter": "org.apache.kafka.connect.storage.StringConverter",
"value.converter": "org.apache.kafka.connect.json.JsonIotaConverter",
```

## Solution

This patch slightly builds upon and expands the existing `default_msg_processor` implementation to accept a few more options to resolve our needs, and to improve matureness closer to what Kafka Connect provides. **The update is fully backwards compatible, which means the patch does not introduce a breaking change for existing users and applications.**

## Details

- Accept a bunch of decoding options per `KafkaDecodingOptions`
- Provide a bunch of output formatting options per `KafkaEvent`
- Tie both elements together using `KafkaEventProcessor`

The machinery is effectively the same like before, but provides a few more options to allow type decoding for Kafka event's key/value slots (`key_type` and `value_type`), a selection mechanism to limit the output to specific fields only (`include`), a small projection mechanism to optionally drill down into a specific field (`select`), and an option to select the output format (`format`).

In combination, those decoding options allow users to relay JSON-encoded Kafka event values directly into a destination table, without any metadata wrappings. Currently, the output formatter provides three different variants out of the box (`standard_v1`, `standard_v2`, `flexible`) [^standard_v2]. More variants can be added in the future, as other users or use cases may have different requirements in the same area.

Most importantly, the decoding unit is very compact, so relevant tasks do NOT require a corresponding **transformation unit** down the pipeline. This keeps the whole ensemble lean, in the very spirit of ingestr.

## Preview
```sh
uv pip install --upgrade 'ingestr @ git+https://github.com/crate-workbench/ingestr.git@kafka-decoder'
```

## Example
Use URL parameters `value_type=json&select=value` to exclusively emit the Kafka event's message payload into the data sink.
```sh
docker run --rm --name=kafka \
  --publish=9092:9092 docker.io/apache/kafka:latest
```
```sh
echo '{"sensor_id":1,"ts":"2025-06-01 10:00","reading":42.42}' | \
  kcat -P -b localhost -t demo
echo '{"sensor_id":2,"ts":"2025-06-01 11:00","reading":451.00}' | \
  kcat -P -b localhost -t demo
```
```sh
ingestr ingest --yes \
  --source-uri "kafka://?bootstrap_servers=localhost:9092&group_id=test&value_type=json&select=value" \
  --source-table "demo" \
  --dest-uri "duckdb:///kafka.duckdb" \
  --dest-table "demo.kafka"
```
```sh
duckdb kafka.duckdb 'SELECT * FROM demo.kafka WHERE sensor_id>1;'
```

## Backlog

- [x] Add software tests for non-standard decoding and output formatting options
- [x] Add docs and improve inline comments

[^standard_v2]: The `standard_v2` output format is intended to resolve GH-289.

### DIFF
--- a/docs/supported-sources/kafka.md
+++ b/docs/supported-sources/kafka.md
@@ -10,29 +10,81 @@ The URI format for Apache Kafka is as follows:
 kafka://?bootstrap_servers=localhost:9092&group_id=test_group&security_protocol=SASL_SSL&sasl_mechanisms=PLAIN&sasl_username=example_username&sasl_password=example_secret&batch_size=1000&batch_timeout=3
 ```
 
-URI parameters:
+### URI parameters
+
+Connectivity options:
 - `bootstrap_servers`: Required, the Kafka server or servers to connect to, typically in the form of a host and port, e.g. `localhost:9092`
 - `group_id`: Required, the consumer group ID used for identifying the client when consuming messages.
 - `security_protocol`: The protocol used to communicate with brokers, e.g. `SASL_SSL` for secure communication.
 - `sasl_mechanisms`: The SASL mechanism to be used for authentication, e.g. `PLAIN`.
 - `sasl_username`: The username for SASL authentication.
 - `sasl_password`: The password for SASL authentication.
+
+Transfer options:
 - `batch_size`: The number of messages to fetch in a single batch, defaults to 3000.
 - `batch_timeout`: The maximum time to wait for messages, defaults to 3 seconds.
+
+Decoding options:
+- `key_type`: The data type of the Kafka event `key` field. Possible values: `json`.
+- `value_type`: The data type of the Kafka event `value_type` field. Possible values: `json`.
+- `include`: A list of event attributes to include in the output, comma-separated.
+- `select`: A single event attribute to select and drill down into.
+  Use `select=value` to relay the Kafka event **payload data** only.
+- `format`: The output format/layout. Possible values: `standard_v1` (default),
+  `standard_v2`, `flexible`. When using the `include` or `select` option, the
+  decoder will automatically select the `flexible` output format.
 
 The URI is used to connect to the Kafka brokers for ingesting messages.
 
 ### Group ID
 The group ID is used to identify the consumer group that reads messages from a topic. Kafka uses the group ID to manage consumer offsets and assign partitions to consumers, which means that the group ID is the key to reading messages from the correct partition and position in the topic.
 
-Once you have your Kafka server, credentials, and group ID set up, here's a sample command to ingest messages from a Kafka topic into a DuckDB database:
+## Examples
 
+### Kafka to DuckDB
+
+Once you have your Kafka server, credentials, and group ID set up,
+here are a few sample commands to ingest messages from a Kafka topic into a destination database:
+
+Transfer data using the traditional `standard_v1` output format into DuckDB.
+The result of this command will be a table in the `kafka.duckdb` database with JSON columns.
 ```sh
 ingestr ingest \
-    --source-uri 'kafka://?bootstrap_servers=localhost:9092&group_id=test_group' \
+    --source-uri 'kafka://?bootstrap_servers=localhost:9092&group_id=test' \
     --source-table 'my-topic' \
-    --dest-uri duckdb:///kafka.duckdb \
+    --dest-uri 'duckdb:///kafka.duckdb' \
     --dest-table 'dest.my_topic'
 ```
 
-The result of this command will be a table in the `kafka.duckdb` database with JSON columns.
+### Kafka to PostgreSQL
+
+Transfer data converging the Kafka event `value` into a PostgreSQL destination
+table, after decoding from JSON, using the `flexible` output format.
+```sh
+echo '{"sensor_id":1,"ts":"2025-06-01 10:00","reading":42.42}' | kcat -P -b localhost -t demo
+```
+```sh
+ingestr ingest \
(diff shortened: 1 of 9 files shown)


docs/supported-sources/kafka.md
```diff
@@ -10,29 +10,80 @@ The URI format for Apache Kafka is as follows:
...
+### Kafka to PostgreSQL
+
+Transfer data converging the Kafka event `value` into a PostgreSQL destination
+table, after decoding from JSON, using the `flexible` output format.
+```sh
+echo '{"sensor_id":1,"ts":"2025-06-01 10:00","reading":42.42}' | kcat -P -b localhost -t demo
+```
+```sh
+ingestr ingest \
+  --source-uri 'kafka://?bootstrap_servers=localhost:9092&group_id=test&value_type=json&select=value' \
```
karakanb, line 68: would the same work without the `select` but the format set to `flexible`?

amotl: Hi. When using `select=value`, the decoder machinery automatically selects the output format `flexible`, you don't need to specify it explicitly.

https://github.com/bruin-data/ingestr/blob/1fc2fc14728fd6f577e04360cdb237d97057e34c/docs/supported-sources/kafka.md?plain=1#L30-L35

Longer explanation: `format={standard_v1,standard_v2,flexible}` is both a parameter and an internal identifier to let the adapter provide three different types of output layouts. This leaves the door open to add other flavours in the future. When the user chooses to use either the `select` or `include` option, the only viable output layout is `flexible`, so the decoder selects it automatically.

docs/supported-sources/kafka.md
```diff
@@ -10,29 +10,80 @@ The URI format for Apache Kafka is as follows:
...
+Decoding options:
+- `key_type`: The data type of the Kafka event `key` field. Possible values: `json`.
+- `value_type`: The data type of the Kafka event `value_type` field. Possible values: `json`.
+- `format`: The output format/layout. Possible values: `standard_v1`, `standard_v2`, `flexible`.
+- `include`: Which fields to include in the output, comma-separated.
+- `select`: Which field to select (pick) into the output.
+
 The URI is used to connect to the Kafka brokers for ingesting messages.
+When using the `include` or `select` option, the decoder will automatically
+select the `flexible` output format.
```
amotl: @karakanb: Does this statement help on the question above you had about the `flexible` output format?

ingestr/src/kafka/model.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if options.format == "standard_v1":
+            UserWarning(
+                "Future versions of ingestr will use the `standard_v2` output format. "
+                "To retain compatibility, make sure to start using `format=standard_v1` early."
+            )
+            return {
+                "_kafka": standard_payload,
+                "_kafka_msg_id": message_id,
+            }
```
amotl, line 134: This section might need further refinement.

a) Is the user warning emitted at all? Let's check that using a test case.
~~b) What if the user decides to deliberately continue using `standard_v1`? Can we have a mechanism to suppress the user warning in this case?~~

docs/supported-sources/kafka.md
```diff
@@ -10,29 +10,80 @@ The URI format for Apache Kafka is as follows:
...
+Transfer options:
 - `batch_size`: The number of messages to fetch in a single batch, defaults to 3000.
 - `batch_timeout`: The maximum time to wait for messages, defaults to 3 seconds.
 
+Decoding options:
+- `key_type`: The data type of the Kafka event `key` field. Possible values: `json`.
+- `value_type`: The data type of the Kafka event `value_type` field. Possible values: `json`.
+- `format`: The output format/layout. Possible values: `standard_v1`, `standard_v2`, `flexible`.
+- `include`: Which fields to include in the output, comma-separated.
+- `select`: Which field to select (pick) into the output.
```
amotl: Todo: Add a remark to the `select` explanation, a compressed version of:

> Always use `select=value` to select the value [sic!] of the Kafka event. This has been chosen deliberately to adhere to the Kafka jargon, despite ingestr's internal event layout relays this field as `data`.

ingestr/src/kafka/model.py
amotl: I think it's okay. Let me know if you think this needs more love.

amotl:
Hi again. Happy New Year! Do you think the patch could be a candidate for merging? Would you like to see any changes or improvements? I've just rebased it on top of `main`.

amotl:
Hi again. I've just refreshed the patch and further improved the documentation.

karakanb:
Hey, we are deprecating the v0 of ingestr, therefore I would like to close this PR. Would you be interested moving over some of the capabilities in this PR into the v1 of ingestr?

amotl:
Hi again,

we are carrying on the Python- and dlt-based v0 codebase with [omniload](https://pypi.org/project/omniload/) and integrated our little improvement with https://github.com/panodata/omniload/pull/51.

Unfortunately, we are not so much fluent with Go, so we can't contribute much, but we will be happy so see corresponding features also being inherited by ingestr v1. 🌻 

With kind regards,
Andreas.

PR #325, opened 2025-07-29
Fix: frankfurter incremental loading

sanjushahgupta:


ingestr/src/sources.py
```diff
@@ -2485,13 +2485,16 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
-            start_date = pendulum.now()
-            end_date = pendulum.now()
+            start_date = pendulum.yesterday()
+
+        if kwargs.get("interval_end"):
+            end_date = ensure_pendulum_datetime(str(kwargs.get("interval_end")))
+        else:
+            end_date = None
+        
+        print("start_date", start_date)
```
turtleDev: we need to remove these debug statements

ingestr/src/frankfurter/helpers.py
```diff
@@ -16,19 +16,19 @@ def get_path_with_retry(path: str) -> StrAny:
...
         raise ValueError("Interval-start cannot be in the future.")
 
     # Check if end_date is in the future
-    if end_date > current_date:
+    if end_date is not None and end_date > current_date:
         raise ValueError("Interval-end cannot be in the future.")
 
     # Check if start_date is before end_date
-    if start_date > end_date:
+    if end_date is not None and start_date > end_date:
```
turtleDev, line 31: You can use a guard clause to reduce repetition.
```python
if end_date is None:
  return
if end_date > current_date:
  ...
if start_date > end_date:
  ...

File: ingestr/src/frankfurter/__init__.py [hunk 3]
```diff
@@ -14,150 +14,152 @@
...
-        initial_value=start_date,
-        end_value=end_date,
-        range_start="closed",
-        range_end="closed",
+    
+
+    @dlt.resource(
+        write_disposition="replace",
+        columns={
+            "currency_code": {"data_type": "text"},
```
turtleDev: `text` is the default data type in dlt anyway. You can remove this and it should be equivalent

PR #338, opened 2025-08-11
Improve docs - adjust, applovin, applovin_max,appsflyer, asana, attio 

sanjushahgupta:


DjamilaBaroudi (approved):

docs/supported-sources/adjust.md
```diff
@@ -55,7 +50,14 @@ Parameters:
 > [!WARNING]
 > Custom tables require a time-based dimension for efficient operation, such as `hour`, `day`, `week`, `month`, or `year`.
 
-## Examples
+| Table           | PK/Merge Key | Inc Key | Inc Strategy | Details                                                                                                                                        |
+| --------------- | ----------- | --------------- | ------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| `Events`        | `id` | –  |        replace     | Retrieves data for [events](https://dev.adjust.com/en/api/rs-api/events/) and event slugs.              |                                        |
+| `campaigns` | `day`| –                | merge            | Retrieves data for a campaign, showing the app's revenue and network costs over multiple days.
```
sabrikaragonen: Can we add which columns are included in this report to the details section? Same for creatives as well.
Also, is there really no inc key?

docs/supported-sources/applovin_max.md
```diff
@@ -45,4 +45,8 @@ By default, `ingestr` retrieves data for the last 30 days. For a custom date ran
...
\ No newline at end of file
+
+Applovin Max source allows ingesting the following sources into separate tables:
+| Table           | PK | Inc Key | Inc Strategy | Details                                                                                                                                        |
+| --------------- | ----------- | --------------- | ------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| [user_ad_revenue](https://developers.applovin.com/en/max/reporting-apis/user-level-ad-revenue-api/)       | partition_date |        partition_date      | merge              |  Provides daily metrics from the user level ad revenue API. User-level revenue data is available eight hours after UTC day end. So, for example, data for UTC 2025-01-01 is available on UTC 2025-01-02 after 08:00.
```
sabrikaragonen: We should define where does partition_date column comes from.

docs/supported-sources/asana.md
```diff
@@ -41,16 +41,17 @@ ingestr ingest \
...
-| `teams`       | Groups of individuals who work together to complete projects and tasks.         |
-| `users`       | Individuals who have access to the Asana platform.                              |
+| Table | Primary/Merge Key | Inc Key | Inc Strategy | Details |
+|-------|----|----------|--------------|---------|
+| `workspaces` | - | - | replace | Information about people, materials, or assets required to complete a task or project successfully. Full reload on each run. |
+| `projects` | - | - | replace | Collections of tasks and related information. Full reload on each run. |
+| `sections` | - | - | replace | Project sections and organization. Full reload on each run. |
+| `tags` | - | - | replace | Labels that can be attached to tasks, projects, or conversations. Full reload on each run. |
+| `tasks` | gid | modified_at | merge | Tasks within a project. Only tasks that belong to a project can be ingested. Uses modified_since API parameter for incremental loading. |
+| `stories` | gid | - | append | Updates or comments that team members can add to a task or project. Depends on tasks, append mode. |
```
sabrikaragonen: Is it really append? Would it not duplicate data?

docs/supported-sources/adjust.md
sanjushahgupta: Ok, I will add the columns. Yes, we don’t have an incremental key because it does not support incremental loading. Though the source implementation says that we support incremental loading but  actually  there is no logic for it, so we need to fix that.

docs/supported-sources/asana.md
sanjushahgupta: It need to be replace and will fix that.

PR #354, opened 2025-08-21
feat/masking on the fly

karakanb:
This PR:
- Adds --mask CLI parameter to apply data masking during ingestion
- Implements 20+ masking algorithms (hash, redact, tokenize, format-preserving, etc.)
- Supports column-specific masking with syntax: --mask "column:algorithm[:param]"

```bash
python -m ingestr.main ingest \
  --source-uri postgresql://localhost/db \
  --source-table users \
  --dest-uri duckdb:///output.db \
  --dest-table masked_users \
  --mask "email:hash" \
  --mask "phone:partial:3" \
  --mask "ssn:redact" \
  --mask "salary:round:1000"
```

image

### DIFF
--- a/docs/getting-started/data-masking.md
+++ b/docs/getting-started/data-masking.md
@@ -1,0 +1,377 @@
+# Data Masking
+
+Data masking is a critical security feature that allows you to protect sensitive information while maintaining data utility for development, testing, and analytics purposes. ingestr provides comprehensive masking capabilities that can be applied to any column during the ingestion process.
+
+## Overview
+
+Data masking transforms sensitive data into a protected format while preserving the structure and type of the original data. This is essential for:
+
+- **Compliance** with regulations like GDPR, CCPA, HIPAA
+- **Security** in development and testing environments
+- **Privacy** protection in analytics and reporting
+- **Data sharing** with third parties or external systems
+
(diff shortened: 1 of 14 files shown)


jbrand-dsp (approved):

docs/getting-started/data-masking.md
```diff
@@ -0,0 +1,377 @@
...
+Data masking is a critical security feature that allows you to protect sensitive information while maintaining data utility for development, testing, and analytics purposes. ingestr provides comprehensive masking capabilities that can be applied to any column during the ingestion process.
+
+## Overview
+
+Data masking transforms sensitive data into a protected format while preserving the structure and type of the original data. This is essential for:
+
+- **Compliance** with regulations like GDPR, CCPA, HIPAA
+- **Security** in development and testing environments
+- **Privacy** protection in analytics and reporting
+- **Data sharing** with third parties or external systems
```
amotl, line 12: Hi. Thanks for the excellent patch. It might make sense to also add a bit of name dropping about [PII](https://en.wikipedia.org/wiki/Personal_data) scrubbing here, because I think it fits into the topic well, closely related to GDPR and HIPAA compliance procedures.

PR #377, opened 2025-09-16
Add elasticsearch destination

mustafa3rsan:


ingestr/src/elasticsearch/helpers.py
```diff
@@ -0,0 +1,115 @@
...
+    with open(file_path, "r") as f:
+        for line in f:
+            if line.strip():
+                doc = json.loads(line.strip())
+                # Clean DLT metadata
+                cleaned_doc = {
+                    k: v for k, v in doc.items() if not k.startswith("_dlt_")
+                }
+                documents.append(cleaned_doc)
+    return documents
```
jbrand-dsp: I would avoid creating and returning a `documents` list as this will consume a lot of memory for large datasets. Instead of `return` use `yield`. E.g.

```
def process_file_items(file_path: str):
    """Process items from a file path (JSONL format) as a generator."""
    with open(file_path, "r") as f:
        for line in f:
            line = line.strip()
            if line:
                doc = json.loads(line)
                # Clean DLT metadata
                cleaned_doc = {k: v for k, v in doc.items() if not k.startswith("_dlt_")}
                yield cleaned_doc
```

ingestr/src/elasticsearch/helpers.py
```diff
@@ -0,0 +1,115 @@
...
+
+def process_iterable_items(items: Any) -> List[Dict[str, Any]]:
+    """Process items from an iterable."""
+    documents = []
+    for item in items:
+        if isinstance(item, dict):
+            # Clean DLT metadata
+            cleaned_item = {k: v for k, v in item.items() if not k.startswith("_dlt_")}
+            documents.append(cleaned_item)
+    return documents
```
jbrand-dsp: Same as comment above

terzioglub (approved):

PR #416, opened 2025-10-20
new source couchbase with sdk

mustafa3rsan:


ingestr/main_test.py
```diff
@@ -560,11 +561,215 @@ def stop_fully(self):
...
+                        return
+            except Exception:
+                pass
+            time.sleep(2)
+
+        raise Exception(f"Bucket '{self.bucket_name}' did not become ready after waiting")
+
+    def _create_primary_index(self):
+        """Create primary index for N1QL queries using Python script inside container."""
+        python_script = f"""
```
karakanb: this should be regular python code

PR #421, opened 2025-10-27
New source hostaway

mustafa3rsan:


ingestr/src/hostaway/client.py
```diff
@@ -0,0 +1,301 @@
...
+            params = {}
+
+        while True:
+            page_params = {**params, "limit": limit, "offset": offset}
+            response = self.session.get(url, params=page_params, timeout=30)
+
+            if response.status_code == 403:
+                break
+
+            response.raise_for_status()
```
karakanb: this is not the right way to check, 403 will silently break out of the loop, we shouldn't break in 403

ingestr/src/sources.py
```diff
@@ -4323,3 +4323,109 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+        if start_date:
+            start_date = ensure_pendulum_datetime(start_date).in_timezone("UTC")
+        else:
+            start_date = pendulum.datetime(1970, 1, 1).in_timezone("UTC")
+
+        end_date = kwargs.get("interval_end")
+        if end_date:
+            end_date = ensure_pendulum_datetime(end_date).in_timezone("UTC")
+        else:
+            end_date = pendulum.now().in_timezone("UTC")
```
karakanb: the `end_date` you used here is supposed to be null if no end date is given, otherwise dlt does not use the state correctly. 

the logic for end_date is:
- if the date is given, filter for those dates
- if the date is not given, use the ones coming from the state

therefore do not make the default `now` here. if you want the default to be `now` if no date is given, then do it in the client

PR #424, opened 2025-11-06
fix: schema with dot; mysql dest_table name

xyxu:
fix: schema with dot
fix: mysql dest_table name


karakanb:
Since we are deprecating the v0 of ingestr I am closing this PR.

PR #427, opened 2025-11-10
New source snapchat ads

mustafa3rsan:
i did not use directly access token because access token expire in an hour , I generate access tokens using refresh tokens

### DIFF
--- a/ingestr/src/snapchat_ads/__init__.py
+++ b/ingestr/src/snapchat_ads/__init__.py
@@ -4,7 +4,6 @@ from typing import Iterator
 
 import dlt
 from dlt.common.typing import TDataItems
-from dlt.sources import DltResource
 
 from .client import SnapchatAdsAPI, create_client
 from .helpers import (


ingestr/src/sources.py
```diff
@@ -4461,3 +4461,211 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+        if resource_name == "event_details":
+            return source.event_details(ad_account_id)
+
+        # Handle creatives specially - it needs ad_account_id parameter
+        if resource_name == "creatives":
+            return source.creatives(ad_account_id)
+
+        # Handle segments specially - it needs ad_account_id parameter
+        if resource_name == "segments":
+            return source.segments(ad_account_id)
```
turtleDev: why not pass `ad_account_id` to source constructor and eliminate this entire block?

ingestr/src/sources.py
```diff
@@ -4461,3 +4461,211 @@ def dlt_source(self, uri: str, table: str, **kwargs):
...
+                    "ad_account_id must be provided in format 'segments:ad_account_id'"
+                )
+        elif table == "segments":
+            # If just "segments" without specific ID, will fetch all ad accounts and their segments
+            resource_name = "segments"
+            ad_account_id = None
+            if not organization_id:
+                raise ValueError(
+                    "organization_id is required for 'segments' table when no specific ad_account_id is provided"
+                )
```
turtleDev: This entire block can be refactored.

Something like this for example:
```python

resources = ["segments", "campagins", ...]
if ":" in table:
    resource_name, account_id = table.split(":",maxsplit=1)
 
 account_id_required = resource_name in resources and account_id is None and organisation_id is None
 if account_id_required:
   raise ValueError(f"organisation id is require for '{resource_name}' when no specific ad_account_id is provided")

File: ingestr/src/snapchat_ads/helpers.py [hunk 3]
```diff
@@ -0,0 +1,257 @@
...
+) -> Iterator[dict]:
+    """
+    Generic helper to fetch data from Snapchat API.
+    """
+    client = create_client()
+    headers = api.get_headers()
+
+    response = client.get(url, headers=headers)
+
+    if response.status_code != 200:
```
turtleDev: use `response.raise_for_status()`

ingestr/src/snapchat_ads/helpers.py
```diff
@@ -0,0 +1,257 @@
...
+) -> Iterator[dict]:
+    """
+    Generic helper to fetch data from Snapchat API with query parameters.
+    """
+    client = create_client()
+    headers = api.get_headers()
+
+    response = client.get(url, headers=headers, params=params or {})
+
+    if response.status_code != 200:
```
turtleDev: use `response.raise_for_status()`

PR #481, opened 2026-01-07
Patch/fireflies api integration

iremcaginyurtturk:
Adds Fireflies.ai as a new source with support for transcripts, analytics, users, channels and other meeting-related data

### DIFF
--- a/ingestr/src/fireflies/__init__.py
+++ b/ingestr/src/fireflies/__init__.py
@@ -14,6 +14,7 @@ def fireflies_source(
     api_key: str,
     start_datetime: Optional[pendulum.DateTime],
     end_datetime: Optional[pendulum.DateTime],
+    granularity: Optional[str] = None,
 ) -> List[DltResource]:
     fireflies_api = FirefliesAPI(api_key=api_key)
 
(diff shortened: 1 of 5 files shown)


mustafa3rsan (commented):
if you want display any docs you need to add it to the ./docs/.vitepress/config.msj

ingestr/src/fireflies/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+    fireflies_api = FirefliesAPI(api_key=api_key)
+
+    start_datetime = (
+        ensure_pendulum_datetime(start_datetime) if start_datetime else None
+    )
+    end_datetime = ensure_pendulum_datetime(end_datetime) if end_datetime else None
+
+    @dlt.resource(
+        write_disposition="replace",
+        primary_key="id",
```
mustafa3rsan: if you are using replace strategy you don't need any primary key or merge key because all tables tables are fully overwritten on each run

ingestr/src/fireflies/helpers.py
```diff
@@ -0,0 +1,770 @@
...
+            yield from self._fetch_analytics_chunk(
+                start.to_iso8601_string(), end.to_iso8601_string()
+            )
+        else:
+            current_start: pendulum.DateTime = start
+
+            while current_start <= end:
+                chunk_end: pendulum.DateTime = current_start.add(days=MAX_DAYS)
+                if chunk_end > end:
+                    chunk_end = end
```
mustafa3rsan, line 487: you can add test for this logic

ingestr/src/fireflies/helpers.py
```diff
@@ -0,0 +1,770 @@
...
+            url=GRAPHQL_API_BASE_URL,
+            json={"query": ACTIVE_MEETINGS_QUERY},
+            headers=self.headers,
+        )
+
+        if response.status_code != 200:
+            error_data = response.json() if response.content else {}
+            raise ValueError(
+                f"Fireflies API Error: {error_data.get('message', 'Unknown error')}"
+            )
```
mustafa3rsan: you can use response.raise_for_status()

PR #601, opened 2026-04-30
Update CrateDB source docs to use PostgreSQL wire protocol

iremcaginyurtturk:
## Summary
- Updated CrateDB source documentation to use PostgreSQL wire protocol (port 5432) instead of HTTP (port 4200)
- Changed source URI scheme from `crate://` to `cratedb://`, aligning source and destination formats
- Simplified destination description by removing references to specific dlt adapters and psycopg2
- Removed unused link references (sqlalchemy-cratedb, dlt cratedb adapter, dlt postgres adapter, psycopg2)

## Test plan
- [ ] Verify documentation renders correctly on the docs site
- [ ] Confirm all link references in the markdown resolve properly

### DIFF
--- a/docs/supported-sources/cratedb.md
+++ b/docs/supported-sources/cratedb.md
@@ -4,41 +4,37 @@
 massive amounts of data in near real-time, even with complex queries. It is
 PostgreSQL-compatible, and based on Lucene.
 
-ingestr supports CrateDB as a source and destination database, using different
-adapters and protocols (HTTP vs. PostgreSQL, see below).
+ingestr supports CrateDB as both a source and destination database.
 
 ## Source
 
-For connecting to CrateDB as a database source, ingestr uses its SQLAlchemy
-dialect package [sqlalchemy-cratedb], effectively talking HTTP to port 4200
-by default.
+For connecting to CrateDB as a database source, ingestr uses the PostgreSQL
+wire protocol on port 5432 by default.
 
 ### URI format
 
 The URI format for CrateDB as a source is as follows:
 ```plaintext
-crate://<username>:<password>@<host>:<port>?ssl=<ssl>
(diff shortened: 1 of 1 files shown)


amotl (commented):
> For connecting to CrateDB as a database source, ingestr uses its SQLAlchemy dialect package [sqlalchemy-cratedb](https://github.com/crate/sqlalchemy-cratedb), effectively talking HTTP to port 4200 by default.

Hi. Thanks for submitting this potential improvement. However, the statement above is still true, so this patch would not reflect reality correctly. We are experimenting with [sqlalchemy-postgresql-relaxed] if that could make the CrateDB dialect talk the PostgreSQL protocol without much ado, but currently the dialect indeed talks HTTP only.

[sqlalchemy-postgresql-relaxed]: https://github.com/pyveci/sqlalchemy-postgresql-relaxed

amotl:
Hi again. After converging the code base to Go, for example when using the pgx driver to connect to CrateDB, this could actually work. Did you observe any problems when connecting to CrateDB using the PostgreSQL protocol, or does it work out of the box?

iremcaginyurtturk:
> Hi again. After converging the code base to Go, for example when using the pgx driver to connect to CrateDB, this could actually work. Did you observe any problems when connecting to CrateDB using the PostgreSQL protocol, or does it work out of the box?

Hi @amotl — yes, the PostgreSQL protocol works out of the box. We just rewrite cratedb:// → postgres:// internally and hand it to pgx on port 5432. Note for users migrating from the Python version: the crate:// alias is no longer accepted (use cratedb://), and SSL is configured via sslmode instead of ssl, as shown in the updated docs. Thanks for the review!

baharkyc (approved):

PR #641, opened 2026-05-26
fix(snowflake): align case-insensitivity and decimal handling for case-folding destinations

iremcaginyurtturk:
## Bug

On non-first runs against a Snowflake target, every non-`replace` strategy produced wrong data. Snowflake stores unquoted identifiers as UPPERCASE; the source emits lowercase; several layers in the pipeline compared names case-sensitively.

Throughout this writeup: **source columns are lowercase** (as the JSONL fixture produces them) and **destination/target table columns are UPPERCASE** (as Snowflake stores them after `CREATE TABLE` via `quoteIdentifier` which uppercases).

---

## `truncate+insert` — every column landed as NULL

**Source** (`JSONL`, 10 rows, lowercase keys):

| id | name | active | score |
|---|---|---|---|
| 1 | alpha | true | 10.5 |
| 2 | bravo | false | 20 |
| … | … | … | … |
| 10 | juliet | false | 100 |

**Target table** *(pre-seeded with 5 rows via `replace`, Snowflake stored as UPPERCASE)* after `truncate+insert`:

| ID | NAME | ACTIVE | SCORE | Before fix |  | ID | NAME | ACTIVE | SCORE | After fix |
|---|---|---|---|---|---|---|---|---|---|---|
| `null` | `null` | `null` | `null` | row 1 |  | 1 | alpha | true | 10.5 | row 1 |
| `null` | `null` | `null` | `null` | row 2 |  | 2 | bravo | false | 20 | row 2 |
| `null` | `null` | `null` | `null` | row 3 |  | 3 | charlie | true | 30.25 | row 3 |
| `null` | `null` | `null` | `null` | … |  | … | … | … | … | … |
| `null` | `null` | `null` | `null` | row 10 |  | 10 | juliet | false | 100 | row 10 |

10 rows landed either way, but **every column was NULL before the fix**.

---

## `delete+insert` — DELETE silently skipped, rows weren't replaced

**Initial seed** (10 rows from JSONL, ids `[1,2,3,4,5,6,8,9,10,11]`):

| id | name |
|---|---|
| 1 | v1-1 |
| 2 | v1-2 |
| 3 | v1-3 |
| … | … |

**Interval source** (5 rows, ids `[3,4,5,6,7]` — strategy should DELETE 3..7 then INSERT these):

| id | name |
|---|---|
| 3 | v2-3 |
| 4 | v2-4 |
| 5 | v2-5 |
| 6 | v2-6 |
| 7 | v2-7 |

**Target table** (Snowflake, UPPERCASE columns) after `delete+insert`:

| ID | NAME (Before fix) | NAME (After fix) |
|---|---|---|
| 1 | v1-1 | v1-1 |
| 2 | v1-2 | v1-2 |
| **3** | **v1-3** ← DELETE skipped, INSERT skipped | **v2-3** |
| **4** | **v1-4** ← stale | **v2-4** |
| 5 | v1-5 | v2-5 |
| 6 | v1-6 | v2-6 |
| **7** | *missing* — never inserted | **v2-7** |
| 8 | v1-8 | v1-8 |
| 9 | v1-9 | v1-9 |
| 10 | v1-10 | v1-10 |
| 11 | v1-11 | v1-11 |
| **Count** | **10** (expected 11) | **11** |

`IntervalTracker.extractValue` had no case for `*array.Decimal128` (Snowflake `NUMBER` incremental key) → returned `nil` → `Min/Max` never set → strategy hit "no interval detected, skip" → DELETE and INSERT both skipped.

---

## `scd2` — CREATE TABLE fails before any data is even read

**Existing target schema** *(returned by `GetTableSchema` — Snowflake stores them UPPERCASE)*:

| Column |
|---|
| `ID` |
| `NAME` |
| `_SCD_VALID_FROM` |
| `_SCD_VALID_TO` |
| `_SCD_IS_CURRENT` |

**The column list `extendSchemaWithSCDColumns` builds for the staging `CREATE TABLE`:**

| Before fix | After fix |
|---|---|
| `ID` | `ID` |
| `NAME` | `NAME` |
| `_SCD_VALID_FROM` *(existing, uppercase)* | `_SCD_VALID_FROM` |
| `_SCD_VALID_TO` *(existing)* | `_SCD_VALID_TO` |
| `_SCD_IS_CURRENT` *(existing)* | `_SCD_IS_CURRENT` |
| **`_scd_valid_from`** ← lowercase constant appended (case-sensitive dedup missed the existing uppercase one) | — |
| **`_scd_valid_to`** ← appended | — |
| **`_scd_is_current`** ← appended | — |

The Snowflake destination's `quoteIdentifier` then uppercases each name when emitting SQL, so the `CREATE TABLE` statement ends up with `"_SCD_VALID_FROM"` listed twice → Snowflake rejects:

```
002025 (42S21): SQL compilation error: duplicate column name '_SCD_VALID_FROM'
```

After the fix, dedup is case-insensitive: the lowercase constant matches the existing uppercase column → not re-appended → `CREATE TABLE` succeeds.

---

## `merge` / `append` — load job fails on NOT NULL primary key

**Source** (JSONL, lowercase):

| id | name | active | score |
|---|---|---|---|
| 1 | alpha | true | 10.5 |
| 2 | bravo | false | 20 |
| … | … | … | … |

**Records the buffer reader emits into staging** *(Arrow field names = staging table column names = UPPERCASE because the buggy `buildBufferReaderTarget` fell into its soft-deleted branch and used the dest-case name)*:

| ID | NAME | ACTIVE | SCORE | Before fix |  | ID | NAME | ACTIVE | SCORE | After fix |
|---|---|---|---|---|---|---|---|---|---|---|
| `null` | `null` | `null` | `null` | row 1 |  | 1 | alpha | true | 10.5 | row 1 |
| `null` | `null` | `null` | `null` | row 2 |  | 2 | bravo | false | 20 | row 2 |
| … | … | … | … | … |  | … | … | … | … | … |

`merge`/`append` then write those rows from staging into the target table where `ID` is `NOT NULL`:

|  | Before fix | After fix |
|---|---|---|
| Snowflake response | `100072 (22000): DML operation to table … failed on column ID with error: NULL result in a non-nullable column` | rows inserted / merged correctly |
| Strategy result | aborts | succeeds |

---

## Test plan

- [x] 7 new unit regression tests fail on `main`, pass with the fix
- [x] Full unit suite (`go test -short ./...`) passes
- [x] Full Snowflake conformance against a real account: 5 previously-failing tests now pass, no regressions
- [x] Full local integration suite passes (Postgres, MySQL, MSSQL, DuckDB, SQLite, CrateDB, CSV, Parquet, Discard)
- [x] Athena conformance against a real AWS account: passes
- [x] BigQuery conformance against a real GCP account: no new failures vs `main`

### DIFF
--- a/pkg/arrowconv/convert.go
+++ b/pkg/arrowconv/convert.go
@@ -580,6 +580,8 @@ func AppendValue(builder array.Builder, val interface{}) {
 			}
 		case *big.Int:
 			b.Append(decimal128.FromBigInt(v))
+		case json.Number:
+			AppendValue(b, string(v))
 		default:
 			b.AppendNull()
 		}
--- a/pkg/arrowconv/convert_test.go
+++ b/pkg/arrowconv/convert_test.go
@@ -283,12 +283,6 @@ func TestAppendValue_TimestampBuilder(t *testing.T) {
 	}
 }
 
-// Regression: Decimal128Builder was the only numeric builder in this file
-// missing a json.Number branch. Schema-inferred sources (JSONL, MongoDB)
-// decode JSON numbers as json.Number via UseNumber(); any DECIMAL/NUMERIC
-// destination column silently received NULL through that path. The
-// delegation to the existing string branch must also inherit the
-// big.Float fallback for scientific notation.
 func TestAppendValue_Decimal128_JSONNumber(t *testing.T) {
 	dt := &arrow.Decimal128Type{Precision: 38, Scale: 0}
 
@@ -296,12 +290,12 @@ func TestAppendValue_Decimal128_JSONNumber(t *testing.T) {
 		name     string
 		val      json.Number
 		wantNull bool
-		wantBigI string // decimal representation expected via BigInt()
+		wantBigI string
 	}{
 		{name: "simple integer", val: json.Number("1"), wantBigI: "1"},
 		{name: "large positive", val: json.Number("42"), wantBigI: "42"},
 		{name: "negative", val: json.Number("-7"), wantBigI: "-7"},
-		{name: "scientific notation (uses big.Float fallback)", val: json.Number("1.5e10"), wantBigI: "15000000000"},
+		{name: "scientific notation", val: json.Number("1.5e10"), wantBigI: "15000000000"},
 		{name: "empty string", val: json.Number(""), wantNull: true},
 		{name: "garbage", val: json.Number("xyz"), wantNull: true},
 	}
--- a/pkg/arrowconv/convert_test.go
+++ b/pkg/arrowconv/convert_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/decimal128"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/bruin-data/ingestr/pkg/schema"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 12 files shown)


iremcaginyurtturk:
@greptileai

karakanb (approved):

PR #672, opened 2026-06-01
Add source_column field to map source→destination column names in ing…

TanayBensuYurtturk:
This pair of PRs adds column name mapping for ingestr assets, allowing columns to be renamed as they move from a source to a destination. In an asset, name stays the destination column and source_column holds the name it has in the source; on ingestion the column is renamed accordingly, with optional type enforcement. The Bruin PR introduces the source_column field and passes the mapping through to ingestr, while the companion ingestr PR handles parsing and applying that mapping on its side. Together they provide a simple, declarative way to standardize column names at load time without post-processing.

### DIFF
--- a/cmd/ingest.go
+++ b/cmd/ingest.go
@@ -148,7 +148,7 @@ func IngestCommand() *cli.Command {
 			},
 			&cli.StringFlag{
 				Name:    "columns",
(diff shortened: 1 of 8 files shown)


TanayBensuYurtturk:
@greptileai

albertobruin (approved):

PR #690, opened 2026-06-02
NetSuite: TBA auth, SuiteAnalytics dialect, and wide-table/CLOB handling

niclas-roos-yubico:
Draft for your review — improvements stacked on top of this connector PR's `netsuite-connector-client` branch. The diff is NetSuite-only (4 files); happy to adjust to fit however you'd like to fold it in.

## What's added on top of the base connector

- **Token-based authentication (TBA).** Computes a per-connection, single-use NetSuite *token password* (HMAC-SHA256 over `accountID&consumerKey&tokenID&nonce&timestamp`, keyed by `consumerSecret&tokenSecret`); a fresh one per physical connection. Works on the DSN and DSN-less driver forms. Username/password still supported.
- **SuiteAnalytics SQL dialect.** `SELECT TOP n` for row limiting (the OpenAccess engine rejects `FETCH FIRST`), interval filtering via `TO_TIMESTAMP`, and no `ORDER BY` (it crashes the driver on wide tables and isn't needed for stateless interval loads).
- **Width-aware column introspection** from the `oa_columns` catalog, so plain table names project columns explicitly instead of `SELECT *` (which fails on very wide tables like `transaction`, ~685 columns).
- **`wide_text` option for wide-table reliability** (see below).

## Wide tables, CLOBs, and an intermittent driver SIGSEGV

On very wide tables the SuiteAnalytics driver intermittently **segfaults** while fetching wide Unicode text columns (`CLOB` + large `VARCHAR2`). Root cause: these land on the `alexbrainman/odbc` wrapper's chunked `SQLGetData` long-data path (`NonBindableColumn`), and the driver faults in its native long-data routine (page-boundary `SEGV_MAPERR`; bundled ICU the likely site). It is **not** the Go wrapper (it passes valid args), and column **ordering does not prevent it**. The fault is amplified when the x86-64 driver runs under emulation (`linux/amd64` on Apple Silicon via QEMU); native x86-64 is reliable.

The new `wide_text` URI option controls those columns:

| value | behaviour |
|-------|-----------|
| `keep` (default) | retrieve in full — correct/lossless on native x86-64 |
| `truncate` | project `SUBSTR(col, 1, N)` (N=`wide_text_max_chars`, default 1000, capped 1024) so the column **binds** and skips the crashing path; keeps all columns, truncates values > N |
| `exclude` | drop the affected columns |

`exclude_clob_columns` is kept for back-compat but only covers `CLOB`-typed columns (15 of the 195 affected on `transaction`); `wide_text` supersedes it.

### Evidence (looping the full incremental load under QEMU)

| projection | chunked-path cols | result |
|---|---|---|
| full / all-wide-last | 195 | SIGSEGV |
| bound columns only | 0 | 5/5 OK |
| wide cols via `SUBSTR(col,1,1000)` | 0 | 5/5 OK |

Verified end-to-end through the connector: `wide_text=truncate` completes 5/5 with all columns.

## Tests

`go test -short ./pkg/source/netsuite/...` — URI/TBA parsing, the SuiteAnalytics query builder, and width-aware classification for each `wide_text` mode plus the catalog fallback.

## Note

The decisive emulation-vs-native control (loop natively on x86-64; valgrind to localize driver vs wrapper) was only run under QEMU here; one native run would confirm the emulation hypothesis. `wide_text` makes wide-table loads reliable in the meantime.


niclas-roos-yubico:
This is actually my first open source PR ever, so please let me know if you need me to change anything!

niclas-roos-yubico:
I'm doing a bit of a rework of this to use the JDBC connector instead, which would introduce JVM as a dependency for anyone wanting to use the Netsuite source, but I think it will make the code better and more stable

PR #692, opened 2026-06-03
feat(salesforce): allow overriding the incremental key

turtleDev:
Salesforce previously rejected --incremental-key and forced a hardcoded replication key per object. Consolidate the SObject name into salesforceTableMeta, honor a user-supplied incremental key as an override (switching to merge), and validate it against the object's datetime fields.

Also only warn about an ignored --incremental-key when a self-managing source actually drops it (resolved key differs from the requested one).

### DIFF
--- a/pkg/source/salesforce/salesforce.go
+++ b/pkg/source/salesforce/salesforce.go
@@ -204,7 +204,7 @@ func (s *salesforceSource) read(ctx context.Context, tableName, replicationKey s
 	go func() {
 		defer close(results)
 
-		sobject := tableName
+		var sobject string
 		if meta, ok := salesforceTableMeta[tableName]; ok {
 			sobject = meta.sobject
 		} else {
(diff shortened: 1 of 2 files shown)


pkg/source/salesforce/salesforce.go
```diff
@@ -121,55 +121,60 @@ func (s *salesforceSource) HandlesIncrementality() bool {
...
 type tableMeta struct {
+	sobject        string
 	strategy       config.IncrementalStrategy
 	pk             []string
 	replicationKey string
 }
 
 var salesforceTableMeta = map[string]tableMeta{
-	"user":                     {config.StrategyReplace, nil, ""},
-	"user_role":                {config.StrategyReplace, nil, ""},
```
TanayBensuYurtturk, line 131: If user gives incremental key for user and user_role tables, there is no `PK` for them. `Id` should be added

TanayBensuYurtturk (approved):

PR #699, opened 2026-06-03
feat: added history tables for existing objects, added case & casehis…

shibba-96:
Added new Salesforce object
- Case

Added history objects for existing Salesforce objects - configured as append only using CreatedDate
- UserHistory
- OpportunityHistory
- AccountHistory
- CaseHistory
- ContactHistory
- LeadHistory
- CampaignHistory


pkg/source/salesforce/salesforce.go
```diff
@@ -255,21 +255,26 @@ type tableMeta struct {
...
 	"campaign_member":          {"CampaignMember", config.StrategyMerge, []string{"Id"}, "SystemModstamp"},
-	"product":                  {"Product2", config.StrategyReplace, []string{"Id"}, ""},
+	"contact":                  {"Contact", config.StrategyReplace, []string{"Id"}, ""},
+	"contact_history":          {"ContactHistory", config.StrategyAppend, []string{"Id"}, "CreatedDate"},
+	"event":                    {"Event", config.StrategyMerge, []string{"Id"}, "SystemModstamp"},
+	"lead":                     {"Lead", config.StrategyReplace, []string{"Id"}, ""},
+	"lead_history":             {"LeadHistory", config.StrategyAppend, []string{"Id"}, "CreatedDate"},
+	"opportunity":              {"Opportunity", config.StrategyMerge, []string{"Id"}, "SystemModstamp"},
+	"opportunity_contact_role": {"OpportunityContactRole", config.StrategyMerge, []string{"Id"}, "SystemModstamp"},
+	"opportunity_history":      {"OpportunityHistory", config.StrategyAppend, []string{"Id"}, "CreatedDate"},
```
karakanb: can you please replace these with "merge" as well so that they can be backfilled if needed without creating duplicates? if the history entries already have ID merge would be preferable.

shibba-96: Updated append to merge

karakanb (approved):

PR #702, opened 2026-06-04
exclude column

TanayBensuYurtturk:
`--sql-exclude-columns `previously matched only the raw source column names, and that check ran before the schema-naming convention renamed columns for the destination. So when a column was renamed (e.g. the default auto/snake_case turns `SecretToken` into `secret_token`), excluding it by the destination name the user actually sees in their warehouse silently did nothing. The column was ingested anyway with no error or warning. This change resolves the naming convention once up front (splitting the old `setupNamingConvention` into `resolveNamingConvention` + `applyNamingConvention`) and reuses it during exclusion, so a column is now dropped from the effective schema if the user names it by either its source name or its post-naming destination name.


TanayBensuYurtturk:
@greptileai

DjamilaBaroudi (approved):

PR #703, opened 2026-06-04
bq partition by date error fix

TanayBensuYurtturk:
Fixes ingesting into BigQuery with `partition_by` set to a column that is already a `DATE` type fails


TanayBensuYurtturk:
@greptileai

terzioglub (approved):

PR #704, opened 2026-06-04
feat(annotation): annotate custom-query and Snowflake source reads

iremcaginyurtturk:
Follow-up to #701 (ETL-phase `type`). Extends `ingestr_extract` coverage to two source reads that #701 didn't reach.

## What

- **Custom query sources** (`source_table: "query: SELECT ..."`) → `ExecuteCustomQuery` now prepends the `ingestr_extract` annotation. A custom query can't use the Storage Read API, so it always runs as a real query → it's annotatable (unlike a plain table source, which reads via the Storage Read API and isn't a query).
- **Snowflake source** reads now set the extract annotation via `QUERY_TAG` (Snowflake strips leading comments), mirroring the Snowflake destination.

## Why

After #701, `ingestr_extract` only appeared for BigQuery **view** sources (SQL fallback). Plain BigQuery table sources (Storage Read API) and Snowflake sources weren't covered. This closes the gap for:
- BigQuery/DuckDB **custom-query** sources (comment in `JOBS.query`)
- **Snowflake** source reads (`QUERY_TAG`)

## Still not annotatable (inherent)
- BigQuery **table** sources → Storage Read API (no query)
- Storage Write API loads → not a query

### DIFF
--- a/pkg/source/adbc/dialect.go
+++ b/pkg/source/adbc/dialect.go
@@ -122,10 +122,3 @@ type URIAware interface {
 	// This is called after successful connection in ADBCSource.Connect().
 	SetURI(uri string)
 }
-
-// SessionQueryTagger is implemented by dialects that carry query attribution
-// via a session QUERY_TAG (Snowflake) instead of a leading comment. The
-// returned statement is run on a pinned connection before a custom query.
-type SessionQueryTagger interface {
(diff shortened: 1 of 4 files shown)


iremcaginyurtturk:
@greptileai

batikankarakan (approved):

PR #707, opened 2026-06-04
fix(postgres): preserve unchanged TOAST/JSONB columns during Postgres CDC merge

### LINKED ISSUE #706: PG CDC merge wipes unchanged TOAST/JSONB columns
On `UPDATE`, logical replication often marks unchanged TOASTed columns as `'u'` (unchanged) in the new tuple.

In pkg/source/postgres_cdc/decoder.go, `parseTupleData` maps `tupleDataUnchanged` → `nil` (“Will use old value if needed”), but `changesToBatch` only reads `change.Values` and never applies `OldValues` (old tuple is only present with REPLICA IDENTITY FULL / O/K in WAL).

Replicated rows therefore have `NULL/empty` for unchanged columns. Snowflake `buildMergeSQL` (and similar methods in other destinations) does `target.col = source.col` for every non-PK column (pkg/destination/snowflake/snowflake.go), which overwrites the warehouse copy.

Genora51:
## Fix: preserve unchanged TOAST/JSONB columns during Postgres CDC merge

Postgres logical replication marks unchanged TOAST columns (large JSONB, text, etc.) with a sentinel byte `'u'` — no payload is sent. Previously the decoder treated `'u'` as `nil`, and destinations used `SET col = source.col`, so any partial UPDATE would wipe large unchanged columns in the warehouse.

### What changed

**Source (`pkg/source/postgres_cdc`)**
- `values.go` — `tupleUnchangedMarker` sentinel distinguishes WAL `'u'` from explicit SQL `NULL`. `resolveColumnValue` falls back to the old tuple when available. `unchangedColumnsJSON` builds a JSON array of column names that were unchanged per UPDATE (e.g. `["config_data"]`). Empty slice is initialised explicitly so `json.Marshal` always produces `[]` rather than `null`.
- New fourth CDC metadata column `_cdc_unchanged_cols` (STRING, `[]` for snapshot/INSERT/DELETE rows).
- `cdc_meta.go` — re-exports canonical column name constants from `pkg/destination`; `sourceColumnCount` uses `len(CDCMetadataColumns())` instead of a hardcoded `-3`.
- Decoder and multitable decoder updated to emit the sentinel and append `_cdc_unchanged_cols`.
- Snapshot rows emit `[]` for the new column.

**Destinations (Postgres, Snowflake, DuckDB, BigQuery, Redshift)**
- Per-column conditional assignment using `_cdc_unchanged_cols`:
  - **Postgres/DuckDB:** `CASE WHEN _cdc_unchanged_cols::jsonb @> '["col"]'::jsonb THEN target.col ELSE source.col END`
  - **Snowflake:** `IFF(ARRAY_CONTAINS(TO_VARIANT('col'), TRY_PARSE_JSON(_cdc_unchanged_cols)), target.col, source.col)`
  - **BigQuery:** `IF('col' IN UNNEST(JSON_EXTRACT_STRING_ARRAY(_cdc_unchanged_cols)), target.col, source.col)`
  - **Redshift:** `CASE WHEN CHARINDEX('["col"', ...) > 0 OR CHARINDEX(',"col"', ...) > 0 THEN target.col ELSE source.col END` (two-anchor check handles multi-element arrays; Redshift has no `jsonb @>`)
- CDC meta columns (`_cdc_lsn`, `_cdc_deleted`, `_cdc_synced_at`, `_cdc_unchanged_cols`) always copy from staging unconditionally.

**Shared (`pkg/destination/cdc_columns.go`)**
- Canonical constants and `CDCMetadataColumns()` / `IsCDCMetaColumn()` — mirrors the `scd2_columns.go` pattern so destinations and the strategy layer share a single source of truth.

### Behaviour

| Case | Result |
|------|--------|
| TOAST column unchanged (`'u'` in WAL) | Listed in `_cdc_unchanged_cols` → target value preserved |
| Explicit `SET col = NULL` | Not listed → staging `NULL` applied correctly |
| Normal value update | Not listed → staging value applied |
| INSERT / DELETE / snapshot | `_cdc_unchanged_cols = []` → all columns written from source |

Fixes #706


turtleDev:
thanks for the contribution @Genora51 

I'll take a look at this later today

Genora51:
Pushed a fix for E2E test failures (duckdb needs a different dialect)

turtleDev:
@Genora51 during testing I see that `_cdc_unchanged_columns` is created as a column in the destination. Do you see a need to keep this column around in the destination?

Genora51:
Oh yeah that's my bad that shouldn't end up in the destination, I'll push a fix

Genora51:
Done

Genora51:
Hey, any update on this? This and #766 are currently blocking my workflow

turtleDev:
hey @Genora51 we've finished testing, but seems like there's a case mismatch bug. We'll merge this and push the fix in a separate PR.

turtleDev:
moving this to https://github.com/bruin-data/ingestr/pull/787

PR #743, opened 2026-06-07
Add MSK IAM OAuth support for Kafka

karakanb:
Adds Kafka OAUTHBEARER support for Amazon MSK IAM using the AWS signer. Documents the new URI parameters and covers the parsing and SASL mechanism paths with Kafka source tests.\n\nValidation: go test ./pkg/source/kafka

### DIFF
--- a/THIRD_PARTY_LICENSES.txt
+++ b/THIRD_PARTY_LICENSES.txt
@@ -9909,6 +9909,194 @@ THE SOFTWARE.
 
 
 ===============================================================================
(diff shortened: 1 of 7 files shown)


slife:
Just noticed this PR. Thanks for putting this together based on the issue report. I was able to push my PR implementation here #752.  However, feel free to use either one as a base. I'm not tied to either one. I'll be happy as long as we get this capability added as an outcome.

PR #744, opened 2026-06-08
refactor(replace): unify primary-key deduplication across destinations

TanayBensuYurtturk:
Replace a per-destination patchwork of dedup logic with a single pre-swap step: when primary keys are set and the destination can merge, load a PK-free raw staging table (staging_raw), dedupe it by PK into a normalised table (staging_normalized) in the target's schema (via MergeTable), then atomically rename it in. Keeping the normalised table in the target schema makes the final swap an O(1) same-schema rename on every destination. Dedup keeps the latest row per key by incremental key, or an arbitrary winner when none is set. Single-table and multi-table replace share the helper.


TanayBensuYurtturk:
@greptileai

turtleDev (approved):

PR #757, opened 2026-06-09
fix(mssql,mysql): use original source column names for read queries

TanayBensuYurtturk:


TanayBensuYurtturk:
Superseded by #758, which re-applies the same fix on a clean branch off current main (avoids the merge conflict with #755) and adds the matching test for the mysql source.

TanayBensuYurtturk:
Superseded by #758 — same fix re-applied on a clean branch off current main (avoids the merge conflict with #755) and adds the matching test for the mysql source.

PR #794, opened 2026-06-11
[codex] Fix Databricks JSON array decoding

### LINKED ISSUE #791: Databricks source Array if int32
I’m back again with another array enhancement, this time for the databricks source connector.

When I try and select a column with an array<int32> I get back an array of NULLs.

I believe this happens because the databricks ExecuteAndWait has a default type of format of JSON_ARRAY

> ... where each value is either the string representation of a value, or null. …

https://docs.databricks.com/api/gcp/workspace/statementexecution/executestatement#format

So even if my type in databricks is of int32, resp.Result.DataArray

Will be a `[[“123”]]` instead of `[[123]]`

So the ‘resp’ in this case https://github.com/bruin-data/ingestr/blob/main/pkg/source/databricks/databricks.go#L420

This means the builder will try and add the int32 value as if it is a string https://github.com/bruin-data/ingestr/blob/main/pkg/source/databricks/databricks.go#L352

This results in a NULL

——

My proof of concept fix which appears to work is:

```
+++ b/pkg/source/databricks/databricks.go
@@ -2,6 +2,7 @@ package databricks
 
 import (
        "context"
+       "encoding/json"
        "errors"
        "fmt"
        "net/url"
@@ -271,6 +272,30 @@ func (s *DatabricksSource) read(ctx context.Context, table string, tableSchema *
        return results, nil
 }
 
+func appendArrayFromJSON(builder array.Builder, val string) {
+       lb, ok := builder.(*array.ListBuilder)
+       if !ok {
+               builder.AppendNull()
+               return
+       }
+
+       var raw []interface{}
+       if err := json.Unmarshal([]byte(val), &raw); err != nil {
+               lb.AppendNull()
+               return
+       }
+
+       lb.Append(true)
+       vb := lb.ValueBuilder()
+       for _, elem := range raw {
+               if elem == nil {
+                       vb.AppendNull()
+               } else {
+                       arrowconv.AppendValue(vb, elem)
+               }
+       }
+}
+
 func (s *DatabricksSource) processResults(ctx context.Context, resp *dbsql.StatementResponse, arrowSchema *arrow.Schema, columns []schema.Column, results chan<- source.RecordBatchResult) {
        batchSize := defaultBatchSize
        alloc := memory.NewGoAllocator()
@@ -348,6 +373,8 @@ func (s *DatabricksSource) buildRecordBatch(alloc memory.Allocator, arrowSchema
                        val := row[i]
                        if val == "" || val == "null" || val == "NULL" {
                                builder.AppendNull()
+                       } else if columns[i].DataType == schema.TypeArray {
+                               appendArrayFromJSON(builder, val)
                        } else {
                                arrowconv.AppendValue(builder, val)
                        }
(END)
```

karakanb:
## Summary

Fixes #791.

Databricks SQL `JSON_ARRAY` statement results return non-null cell values as strings. For `ARRAY<INT>` columns, that means the source receives a JSON-encoded array cell such as `[123,"456",null]`. The shared Arrow list append path previously treated string inputs as invalid list values and appended `NULL` for the whole array.

This updates the Arrow conversion layer to parse string inputs as JSON arrays when writing into a list builder, preserving numeric values with `json.Number`, and adds Databricks regression coverage for integer and string array cells. It also preserves empty string cells from Databricks as empty values while keeping `"null"`/`"NULL"` sentinel values as nulls.

## Validation

- `go test ./pkg/arrowconv ./pkg/source/databricks`
- `make format`
- `make lint`
- `make test`
- Live Databricks probe with `SELECT array(123, 456, CAST(NULL AS INT)) AS ints, array('alpha', 'beta') AS strings`

### DIFF
--- a/pkg/arrowconv/convert.go
+++ b/pkg/arrowconv/convert.go
@@ -727,6 +727,15 @@ func AppendValue(builder array.Builder, val interface{}) {
 }
 
 func appendListValue(b *array.ListBuilder, val interface{}) {
+	if s, ok := val.(string); ok {
+		values, ok := parseJSONStringArray(s)
+		if !ok {
+			b.AppendNull()
+			return
+		}
+		val = values
+	}
+
 	rv := reflect.ValueOf(val)
 	if rv.Kind() != reflect.Slice {
 		b.AppendNull()
@@ -743,6 +752,27 @@ func appendListValue(b *array.ListBuilder, val interface{}) {
 		}
 		AppendValue(vb, elem)
 	}
+}
+
+func parseJSONStringArray(s string) ([]interface{}, bool) {
+	s = strings.TrimSpace(s)
+	if s == "" {
+		return nil, false
+	}
+	if s[0] != '[' {
+		return nil, false
+	}
+
+	var values []interface{}
+	dec := json.NewDecoder(strings.NewReader(s))
+	dec.UseNumber()
+	if err := dec.Decode(&values); err != nil {
+		return nil, false
+	}
+	if dec.More() {
+		return nil, false
+	}
+	return values, true
 }
 
 func listElementValue(v reflect.Value) (interface{}, bool) {
--- a/pkg/source/databricks/databricks.go
+++ b/pkg/source/databricks/databricks.go
@@ -346,9 +346,12 @@ func (s *DatabricksSource) buildRecordBatch(alloc memory.Allocator, arrowSchema 
 				continue
 			}
 			val := row[i]
-			if val == "" || val == "null" || val == "NULL" {
+			switch val {
+			case "":
+				builder.AppendEmptyValue()
+			case "null", "NULL":
 				builder.AppendNull()
-			} else {
+			default:
 				arrowconv.AppendValue(builder, val)
 			}
 		}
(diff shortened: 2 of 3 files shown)


markdessain:
While we are here can we also add.

Just as a “” string should be a “” an

```
diff --git a/pkg/source/databricks/databricks.go b/pkg/source/databricks/databricks.go
index 04469c87..91762fb1 100644
--- a/pkg/source/databricks/databricks.go
+++ b/pkg/source/databricks/databricks.go
@@ -2,6 +2,7 @@ package databricks
 

 func (s *DatabricksSource) processResults(ctx context.Context, resp *dbsql.StatementResponse, arrowSchema *arrow.Schema, columns []schema.Column, results chan<- source.RecordBatchResult) {
        batchSize := defaultBatchSize
        alloc := memory.NewGoAllocator()
@@ -346,8 +373,16 @@ func (s *DatabricksSource) buildRecordBatch(alloc memory.Allocator, arrowSchema
                                continue
                        }
                        val := row[i]
-                       if val == "" || val == "null" || val == "NULL" {
+                       if val == "" {
+                               builder.AppendEmptyValue()
+                       } else if val == "null" || val == "NULL" {
```

iremcaginyurtturk (approved):

albertobruin (approved):

PR #797, opened 2026-06-14
Add --stream flag for continuous (streaming) ingestion

karakanb:
## What

Adds a first-class `--stream` flag that turns one-shot `ingest` into a long-running flush loop. Supported by CDC sources (`postgres+cdc`, `mssql+cdc`) and message brokers (`kafka`, `amqp`). Records are buffered and flushed to the destination whenever `--flush-interval` (default `30s`) **or** `--flush-records` (default `50000`) is reached — whichever comes first — until interrupted (SIGINT/SIGTERM), which triggers a final flush and clean exit.

## How it works

- **Capability interfaces** (`pkg/source/source.go`): `StreamingSource` gates the flag and picks the default strategy; `StreamCommitter` provides durability feedback. `RecordBatchResult` gains a cumulative `CommitToken`; `ReadOptions`/`TableRequest` gain `Streaming`.
- **Streaming executor** (`pkg/strategy/streaming.go`): buffers Arrow batches per table and runs each cycle as `write → merge → reset-staging → commit`. This ordering gives **at-least-once** delivery — a crash before commit re-delivers, and the merge (by PK / `msg_id`) makes replays idempotent. Memory is bounded by `--flush-records`.
- **Durability:** Postgres CDC confirms the replication-slot LSN only after a flush+merge; RabbitMQ acks and Kafka commits group offsets only after a flush. MSSQL CDC resumes from the destination's `max(_cdc_lsn)` (no consumer ack concept).
- **Brokers** stream into a fixed envelope: `msg_id` (PK) + JSON `data` + `_ingestr_order` (Kafka offset / AMQP delivery tag). Merged on `msg_id` with last-value-per-key ordering within each flush window. Append mode does not enforce the PK (so at-least-once redeliveries don't crash on duplicate keys).

## Bugs fixed during validation

- **Multi-table CDC boundary data loss:** the per-table filter used `changeLSN <= processedLSN`, dropping the first transaction whose BEGIN LSN equals the snapshot consistent-point (it is *not* in the snapshot). Changed to strict `<` (safe on resume — merge is idempotent).
- **Replication read stall:** a 100ms per-call `ReceiveMessage` deadline churned the connection and raced pgconn's background reader, wedging the stream after the first flush under load. Raised to 1s and added dead-connection detection (fatal standby-send failure + a 2-minute liveness timeout) so a broken connection errors out for a supervisor to restart instead of hanging silently.

## Testing

- Unit tests: flush loop (count/interval triggers, empty-cycle skip, commit ordering, abort, graceful shutdown, routing, truncate fallback, append-no-PK, incremental-key ordering), `safeCommitLSN`, commit-state, config validation.
- Integration tests (`//go:build integration`): Postgres CDC streaming + resume, RabbitMQ, Kafka (adds a Redpanda testcontainer), MSSQL CDC.
- Manually validated end-to-end: 10M+ messages Kafka→Postgres with exact counts, bounded memory, and clean shutdown.

## Notes for reviewers

- The 100ms→1s receive timeout also applies to the existing **batch-mode** CDC path (idle-poll cadence only; batch tests pass) — intentional, it's strictly less wasteful.
- New dependency `testcontainers-go/modules/redpanda` is **test-only**.

### DIFF
--- a/cmd/ingest.go
+++ b/cmd/ingest.go
@@ -181,6 +181,23 @@ func IngestCommand() *cli.Command {
 				Usage:   "Enable debug logging",
 				Sources: cli.EnvVars("DEBUG", "INGESTR_DEBUG"),
 			},
+			&cli.BoolFlag{
+				Name:    "stream",
+				Usage:   "Continuously ingest from the source, flushing to the destination on an interval or record-count trigger (CDC and message-broker sources only)",
+				Sources: cli.EnvVars("INGESTR_STREAM"),
+			},
+			&cli.DurationFlag{
+				Name:    "flush-interval",
+				Usage:   "How often to flush buffered records to the destination in streaming mode",
+				Value:   30 * time.Second,
+				Sources: cli.EnvVars("INGESTR_FLUSH_INTERVAL"),
+			},
+			&cli.IntFlag{
+				Name:    "flush-records",
+				Usage:   "Flush to the destination when this many records are buffered in streaming mode",
+				Value:   50000,
+				Sources: cli.EnvVars("INGESTR_FLUSH_RECORDS"),
+			},
 			&cli.StringFlag{
 				Name:    "query-annotations",
 				Usage:   "JSON object of caller annotation keys (e.g. {\"pipeline\":\"p\",\"asset\":\"a\"}) merged into the '-- @bruin.config' comment on destination queries (QUERY_TAG on Snowflake) for cost attribution. ingestr always annotates with its own keys (type, ingestr_step); this flag adds the caller's keys on top.",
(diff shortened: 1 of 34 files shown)


karakanb:
Thanks @greptileai — good catch on the graceful-shutdown commit. Fixed in d7344e08.

Both Kafka and RabbitMQ had the same lifecycle bug: the reader/channel was torn down in the read goroutine on cancellation, before the pipeline's final flush ran, so `CommitStream` saw a nil reader/channel and errored. A SIGINT with un-flushed data would exit with an error instead of cleanly (the existing tests only cancelled after the stream had fully caught up, so they didn't exercise it).

Rather than the suggested no-op alone (which would re-deliver the last flush window on every restart), I kept the reader/channel **open past its goroutine** so the final flush can still commit/ack, and moved the close into `Close()` (which the pipeline calls after `Run` returns). `CommitStream` *also* treats a nil reader/channel as a safe no-op as a defensive fallback.

Added `TestRabbitMQ_StreamingShutdownFlushesPending`: cancels with buffered, un-flushed data and asserts a clean exit, that the final flush wrote all rows, the queue is drained (acked), and a re-run sees no redelivery. Kafka + RabbitMQ streaming integration tests and the full unit suite pass.

karakanb:
Fixed in the latest commit — and a real data-loss bug, thanks @greptileai.

I scoped the fix slightly differently than suggested: `generateMsgID` is kept as-is because the non-streaming `_kafka_msg_id` metadata column is Python-compatible (`hash(topic+partition+"None")` for keyless), asserted by `TestGenerateMsgID_NilKey`. Changing it there would break that contract.

Instead, the **streaming envelope's primary key** now comes from a new `streamMsgID`:
- **Keyed topics:** key-based id (offset ignored) → same key dedups to the latest record (changelog/compaction semantics, unchanged).
- **Keyless topics:** `topic:partition:offset` → every message gets a distinct id, so nothing collapses under merge. It's stable across redeliveries (offset is fixed), so at-least-once dedup still holds.

Added `TestStreamMsgID` covering both cases. Kafka streaming integration test + unit suite pass, lint clean.

turtleDev (approved):

PR #813, opened 2026-06-17
Add pyarrow-hotfix to mitigate CVE-2023-47248 on older pyarrow

turtleDev:
Adds pyarrow-hotfix to the [sdk] extra and imports it in _require_pyarrow(). It neutralizes CVE-2023-47248 on pyarrow <14.0.1 (e.g. the 12.0.1 resolved for Python 3.7) and is a no-op on patched versions, so no minimum Python/pyarrow version is forced.


batikankarakan (approved):

turtleDev:
I realised that we don't ready data via pyarrow (which is the vulnerable path). This change is not necessary.

PR #817, opened 2026-06-18
Add SharePoint source

krom-als:
Adds a `sharepoint` source that reads Excel and CSV files from a SharePoint Online document library through the Microsoft Graph API.

A connection points at a single site, and the source table is the path to a file within that site's document library, optionally followed by a few hints — which sheet or sheets to read, how many rows to skip, the CSV encoding and separator, and which columns hold dates. Paths can use glob patterns to pull many files at once, and multiple Excel sheets can be stacked into one table, matched up by column name. Every row is landed as raw text along with three extra columns (`_source_file`, `_sheet_name`, `_row_idx`) that record where each row came from, so the originating file and the original row order are preserved.

Files are read in a streaming fashion, so a large multi-sheet workbook stays in the few-hundred-MB range instead of being loaded whole. Excel dates are kept as their raw serial numbers by default and only turned into ISO strings for columns named in a `date_cols` hint, which keeps the read fully streaming since it avoids per-cell style lookups. Legacy `.xls` files aren't supported and produce a clear error suggesting `.xlsx`.

Authentication is the OAuth2 client-credentials flow, so the app registration just needs read access to the site. The source supports the standard write strategies and defaults to replace. It does not do incremental extraction — each run reads the whole file or glob.

A quick example:

```
ingestr ingest \
  --source-uri 'sharepoint://?tenant_id=...&client_id=...&client_secret=...&hostname=example.sharepoint.com&site=sites/Example' \
  --source-table 'Reports/budget.xlsx#sheets=North|South,date_cols=Date' \
  --dest-uri duckdb:///out.db --dest-table raw.budget
```

I tested it end to end against a real SharePoint site loading into both Snowflake and DuckDB, comparing the output against an existing pipeline — exact parity on the simple tables and matching values on the larger multi-sheet and globbed ones. There are unit tests covering the source-table parsing, globbing, header and blank-row handling, the multi-sheet column union, date conversion, and CSV streaming, plus an integration test gated behind credentials.

A couple of things left as follow-ups: honoring the Graph `Retry-After` header when the API throttles, and parallelizing file downloads across a large glob.

### DIFF
--- a/docs/.vitepress/config.mjs
+++ b/docs/.vitepress/config.mjs
@@ -220,6 +220,7 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
               { text: "S3", link: "/supported-sources/s3.md" },
               { text: "Salesforce", link: "/supported-sources/salesforce.md" },
               { text: "SFTP", link: "/supported-sources/sftp.md"},
+              { text: "SharePoint", link: "/supported-sources/sharepoint.md" },
               { text: "Shopify", link: "/supported-sources/shopify.md" },
               { text: "Slack", link: "/supported-sources/slack.md" },
               { text: "Smartsheet", link: "/supported-sources/smartsheets.md" },
--- a/docs/supported-sources/sharepoint.md
+++ b/docs/supported-sources/sharepoint.md
@@ -20,8 +20,13 @@ URI parameters:
 - `hostname`: SharePoint tenant hostname, e.g. `example.sharepoint.com`
 - `site`: server-relative site path, e.g. `sites/Example`
 - `library` (optional): document library (drive) name; defaults to the site's default **Documents** library. Set it to read from a secondary library on the same site.
+- `max_file_size` (optional): maximum bytes for any single downloaded file; the read fails if a file exceeds it. Defaults to `0` (unlimited).
+- `max_files` (optional): maximum number of files a glob may match before erroring, to guard against an over-broad recursive pattern. Defaults to `10000`; set `0` for unlimited.
 
 Authentication uses the OAuth2 **client-credentials** flow. The app registration needs application permission to read the site's files (e.g. `Sites.Read.All` or `Files.Read.All`), granted with admin consent.
+
+> [!TIP]
+> The `client_secret` travels in the URI, which can end up in shell history or process listings. Prefer supplying it from a secret store or environment variable rather than hard-coding it.
 
 To read files from a different site (a different document library), define a second connection pointing at that `site`.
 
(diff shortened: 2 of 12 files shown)


karakanb:
@greptileai

arsalann:
Security/code review findings from a local pass against `origin/main`:

1. **P1: Add bounded download and workbook parse limits.** `downloadToFile` streams Graph content to disk with an unbounded `io.Copy`, and `readExcel` opens workbooks with Excelize defaults. A huge or malicious SharePoint file can exhaust disk, memory, or CPU before row streaming helps. Please add a maximum downloaded file size and pass explicit `excelize.Options{UnzipSizeLimit, UnzipXMLSizeLimit, TmpDir}` when opening `.xlsx` files. Relevant areas: `pkg/source/sharepoint/graph.go` around `downloadToFile`, and `pkg/source/sharepoint/excel.go` around `excelize.OpenFile`.

2. **P2: Bound recursive glob enumeration.** `**` enables recursive traversal and the implementation accumulates every match before reading. A broad pattern such as `**/*.xlsx` can enumerate a whole document library, burn Graph quota, and grow memory unexpectedly. Please add limits such as max pages/files/depth with a clear error, or stream matches instead of collecting the full list first. Relevant area: `pkg/source/sharepoint/graph.go` around `listMatching` / `walk`.

Risk notes:
- `govulncheck ./pkg/source/sharepoint` reports reachable issues from the local Go toolchain and existing `golang.org/x/net`; those appear mostly pre-existing versus `origin/main`, but release builds should use patched Go and dependency versions.
- `excelize v2.10.1` brings additional indirect dependencies. Current `x/image` advisories do not look reachable in the SharePoint import graph, but module-level scanners may still flag them.
- The docs use `client_secret` in URI examples. That matches existing ingestr URI patterns, but it would be safer to steer users toward env vars to avoid shell history/process-list exposure.

Checks run locally:
- `git diff --check origin/main...pr-817-review`: passed
- `go test ./pkg/source/sharepoint`: passed
- `go test -short ./pkg/source/sharepoint`: passed
- `make build`: passed
- `make lint`: `go vet` passed, then failed because `golangci-lint` is not installed in this environment

I did not run a live SharePoint ingestion test.

krom-als:
Pushed a commit addressing the review feedback:

- **Header dedup collision** — a generated suffix that matched an existing column name could produce two columns with the same name (dropping one); it now keeps incrementing until the name is unique.
- **Globs without an extension** — instead of erroring with "could not determine file format", each matched file's format is now detected individually, so one glob can span mixed file types.
- **Workbook parse limits** — Excel reads now pass explicit excelize unzip limits and a temp dir as a decompression-bomb guard; large-file streaming is unchanged.
- **Optional bounds** — added `max_file_size` (per-file download cap) and `max_files` (glob match cap, default 10000) URI params.
- **Docs** — documented the new params and added a note to prefer supplying the secret from a secret store / env var rather than inline.

Still to do: regenerate the third-party license notices for the new transitive dependencies (the failing `licenses` check).

karakanb (approved):

PR #825, opened 2026-06-19
fix licenses test

iremcaginyurtturk:
run command: `make licenses`


iremcaginyurtturk:
Moving the THIRD_PARTY_LICENSES.txt change into #823 instead.

PR #917, opened 2026-07-02
Fix Snowflake multi-table write deadlock from connection pool exhaustion

### LINKED ISSUE #916: Snowflake connection pool can deadlock for multi-table batch
Snowflake destination connections are limited to a pool of 16. For multi-table sync, ingestr creates 5 worker threads per table, and each thread holds onto a snowflake connection for its lifetime. If the number of workers exceeds 16, the remaining workers get stuck forever. If all 5 workers for a particular table lose this race and get stuck, then ingestr hangs.

Genora51:
## Summary

Multi-table Snowflake writes (e.g. Postgres CDC with `source_table: "*"` and `incremental_strategy: merge`) could hang indefinitely with zero `PUT`/`COPY` activity once `numTables * ExtractParallelism` exceeded the Snowflake connection pool size (`SetMaxOpenConns(16)`).

**Root cause:** `WriteParallel` acquired a dedicated `*sql.Conn` per upload worker *before* the worker touched its input channel, and held that connection for the worker's entire lifetime — including while idle. With enough tables/workers, some workers permanently starved the pool while sitting idle waiting for data, and the multi-table `Router`'s single goroutine would then block forever trying to deliver to a table whose workers could never get a connection — a circular deadlock with no error surfaced.

**Fix:** stop pinning a connection per worker. Each batch's `PUT` now goes through `d.db.ExecContext(uploadCtx, putSQL)` directly, so `database/sql` acquires a connection from the shared pool only for the duration of that single statement and releases it immediately after. This lets the existing pool be safely time-shared across any number of tables/workers without needing to cap parallelism based on table count.

Fixes #916.

## Changes

- `pkg/destination/snowflake/snowflake.go`: removed the per-worker `db.Conn()`/`defer conn.Close()`; `PUT` uploads now use `d.db.ExecContext` per batch.
- `pkg/destination/snowflake/snowflake_test.go`: added `TestMultiTableWriteDoesNotDeadlockUnderConnectionPressure`, an end-to-end repro using the real `multitable.Write`/`Router` with a `sqlmock`-backed pool smaller than `numTables * Parallelism` — verified this fails/times out against the old code and passes with the fix.
- `pkg/destination/multitable/writer_test.go`: added `TestWriteDoesNotDeadlockWithScarceSharedResourceScopedPerBatch`, a generic regression test documenting that the router design is safe as long as a scarce shared resource is only held per-batch, never while idle.

### DIFF
--- a/pkg/destination/multitable/writer_test.go
+++ b/pkg/destination/multitable/writer_test.go
@@ -3,6 +3,7 @@ package multitable
 import (
 	"context"
 	"errors"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


karakanb (approved):

karakanb:
@greptileai

PR #928, opened 2026-07-07
feat(trello): add Trello source

iremcaginyurtturk:
## Summary
Adds a **Trello** source that ingests data from the Trello REST API.

**URI:** `trello://?api_key=<key>&token=<token>` (both passed as query params on every request; the key + token come from a Trello Power-Up).

## Tables
All tables use `id` as the primary key and infer their schema (nested objects land as JSON columns; only `id` is lifted to the top level).

| Table | Endpoint | Strategy | Incremental key | Filtering |
|---|---|---|---|---|
| boards | `/members/me/boards` | replace | – | none |
| organizations | `/members/me/organizations` | replace | – | none |
| lists | `/boards/{id}/lists` | replace | – | none |
| members | `/boards/{id}/members` (deduped by id) | replace | – | none |
| labels | `/boards/{id}/labels` | replace | – | none |
| checklists | `/boards/{id}/checklists` | replace | – | none |
| cards | `/boards/{id}/cards` | merge | `dateLastActivity` | client-side |
| actions | `/boards/{id}/actions` | merge | `date` | client-side on `max(date, dateLastEdited)` |

### DIFF
--- a/README.md
+++ b/README.md
@@ -684,6 +684,11 @@ Pull requests are welcome. However, please open an issue first to discuss what y
         <td>-</td>
     </tr>
     <tr>
+        <td>Trello</td>
+        <td>✅</td>
+        <td>-</td>
+    </tr>
+    <tr>
         <td>Trustpilot</td>
         <td>✅</td>
         <td>-</td>
--- a/docs/.vitepress/config.mjs
+++ b/docs/.vitepress/config.mjs
@@ -240,6 +240,7 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
               { text: "Stripe", link: "/supported-sources/stripe.md" },
               { text: "SurveyMonkey", link: "/supported-sources/surveymonkey.md" },
               { text: "TikTok Ads", link: "/supported-sources/tiktok-ads.md" },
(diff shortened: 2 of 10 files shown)


iremcaginyurtturk:
Addressed the Greptile review:

1. **Registration unreachable (P1)** — Not an issue. `internal/registry/imports/imports.gen.go` is generated (gitignored) and every `make build`/`test`/`test-integration` target depends on `generate`, which now includes `pkg/source/trello`. `trello://` resolves — the integration test (which blank-imports via the generated file) and all live runs confirm it. Same mechanism as every other source.

2. **Fan-out batches infer different schemas (P2)** — By design. The pipeline feeds every board batch into a single schema inferrer and buffers them, merging into one superset schema before writing (`inferSchemaFromData`). `nil` columns per batch is the prescribed pattern for inference sources; heterogeneous boards merge to a union, same as `attio`/`clickup`.

3 & 4. **Pagination misses concurrent inserts (P2)** — Inherent to `before`-cursor pagination on live data and identical to Airbyte's Trello connector. Items created mid-scan are newer than every subsequent cursor, so they're picked up on the next incremental run; with `merge` on `id` nothing is lost over time.

DjamilaBaroudi (dismissed):

DjamilaBaroudi (approved):

PR #930, opened 2026-07-07
docs: connector-catalog registration + return-everything params for new sources

iremcaginyurtturk:
Two additions to the "Adding New Sources" guidance in `CLAUDE.md` (and `AGENTS.md`, which symlinks to it), based on gaps hit while adding the Trello source:

1. **Register the scheme in the server connector catalog.** `internal/server/connectors.go` (`GetConnectors()`) is a separate, hand-maintained list from the URI registry. Missing it makes `TestConnectorCatalogCoversRegisteredSchemes` fail with `scheme "<x>" is missing from connector catalog`.

2. **Check "return everything" query params.** Many vendor endpoints default to only active/open/visible items (and a limited type subset), silently dropping archived/closed/deleted records. Sources should pass the param that returns the full set — e.g. Trello `filter=all`, or `include_archived=true` / `state=all` / `deleted=true` elsewhere — and verify empirically.

Docs-only; no code changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -176,10 +176,13 @@ The tool accepts various URI schemes:
 ### Adding New Sources
 1. Implement the `Source` interface or create an ADBC Dialect
 2. Register in `internal/uri/registry.go` `init()` with URI schemes
-3. If using ADBC: implement Dialect with SQL templates and type mapper
-4. If native driver: handle connection, schema fetching, and batch reading directly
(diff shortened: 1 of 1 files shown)


iremcaginyurtturk:
Superseded — the source-writing guidance lives in the add-source skill (.agents/skills/add-source/SKILL.md), not CLAUDE.md. Reopened as the SKILL.md update in the new PR.

Issue #926, opened 2026-07-07
OneLake ingestr destination does not support schema-enabled Fabric Lakehouses


crazy-treyn:
## Summary

Bruin's ingestr operator can't land data into schema-enabled Fabric lakehouses, because it hard-codes the ingestr `--dest-table` to the asset name and offers no way to override it.

## Environment

- bruin `v0.11.666`
- ingestr `1.0.66`
- `onelake` destination

## Details

Fabric lakehouses with schema support store tables at `Tables/<schema>/<table>`. ingestr's OneLake destination handles this correctly when given `--dest-table "Tables/<schema>/<table>"` — it writes a proper Delta table that the SQL endpoint registers under the schema (verified).

However, Bruin's ingestr operator sets `destTable := asset.Name` in `pkg/ingestr/operator.go` with no override, and asset names must satisfy `task-name-valid` (alphanumerics, dashes, dots, underscores — **no `/`**). So a dotted asset name like `myschema.mytable` is passed through verbatim and lands as a *flat* folder literally named `myschema.mytable` under `Tables/`, which a schema-enabled lakehouse does not register as a table.

karakanb:
We are taking a look at this.

turtleDev:
hey @crazy-treyn would it make sense as a one lake user if Bruin translates `schema.table` to `schema/table` internally?

For instance
```
name: raw.sales
type: ingestr
connection: onelake
parameters:
  source_table: staging.raw_sales
  source_connection: upstream
```

would get translated to
```
ingestr ingest ... --dest-table "Tables/raw/sales"
```

crazy-treyn:
@turtleDev Yeah that would make sense. 

Since we are somewhat used to the slash when in Onelake and dealing abfss paths, it would be good to have an example in the docs of the ingestr table config for a schema-enabled lakehouse, as well as a non schema-enabled lakehouse.

turtleDev:
hey @crazy-treyn we made ingestr alias `schema.prefix` to `schema/prefix` automatically. Support for this should land in `v1.0.70`.

PR #934, opened 2026-07-07
onelake: accept period-separated Delta table targets

turtleDev:
## What

The OneLake destination previously accepted only slash-separated `--dest-table` values (`Tables/schema/name`). This adds periods as an equivalent separator for Delta table targets:

| `--dest-table` | Resolves to |
|---|---|
| `name` | `Tables/name` |
| `schema.name` | `Tables/schema/name` |
| `Tables.schema.name` | `Tables/schema/name` |
| `Tables/schema/name` (existing) | `Tables/schema/name` |

`.` and `/` are interchangeable separators and the leading `Tables` segment is optional.

## Why

Period-separated names are a natural way to express `schema.table` and are unambiguous here because Fabric table and schema names cannot contain a period.

## Files mode

Unchanged. Files targets still require the explicit `Files/<path>` slash form, and the remainder is taken verbatim (no period conversion) so file extensions such as `data.parquet` are preserved.

## Scope notes

- All existing callers (`tableDir`, `tableDirForTables`, `DropTable`) inherit the new parsing via `parseTarget`.
- No deep-nesting validation: Fabric supports a single schema level, but `Tables.a.b.c` maps literally to `a/b/c` rather than erroring.
- A schema literally named `tables` (`tables.x`) collapses to `Tables/x`, consistent with the existing `Tables/` prefix handling.

Tests cover the new period forms and lock the no-conversion behavior for Files. Docs updated.

### DIFF
--- a/docs/supported-sources/onelake.md
+++ b/docs/supported-sources/onelake.md
@@ -25,6 +25,8 @@ The **mode and table** come from `--dest-table`, mirroring OneLake's path layout
 - `Files/<path>` → raw Parquet files
 - a bare name with no prefix defaults to `Tables/`
 
+For Delta tables, `.` and `/` are interchangeable separators and the leading `Tables` segment is optional, so `users`, `schema.name`, `Tables.schema.name` and `Tables/schema/name` are all valid (the last two are equivalent). Period-separated names are convenient because Fabric table and schema names cannot contain a period. Files targets must use the explicit `Files/<path>` form — periods there are left untouched so file extensions are preserved.
+
 The final object path is:
 `https://onelake.dfs.fabric.microsoft.com/<workspace>/<lakehouse>.Lakehouse/<Tables|Files>/<rest>`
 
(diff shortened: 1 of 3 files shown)


TanayBensuYurtturk (dismissed):

turtleDev:
Thanks @greptile-apps — both addressed in 5b45507f:

- **P1 (bare `Tables` deletes root):** `parseTarget` now returns an error, and `DropTable` propagates it, so a bare `Tables` target (empty relative path) is rejected instead of resolving to the top-level `Tables` area. The guard is centralized in `parseTarget` (previously duplicated in `PrepareTable`/`tableDirForTables` and missing in `DropTable`).
- **P2 (empty segments collapse):** switched from `FieldsFunc` to split-keeping-empties + validation, so `schema..name` / `.users` are now rejected rather than silently normalized.

turtleDev:
closes https://github.com/bruin-data/ingestr/issues/926

DjamilaBaroudi (approved):

TanayBensuYurtturk (approved):

PR #948, opened 2026-07-09
feat(bigquery): reconcile partition/cluster changes and normalize partition names

iremcaginyurtturk:
## What 

`SwapTable` now drops a target whose partition/cluster spec differs from the requested one before recreating it. This is safe because `SwapTable` is only reached from the **replace** strategy, which fully overwrites the target anyway — so no data is lost beyond what replace already discards, and merge/append never reach this code.

New helper `partitionOrClusterMismatch` compares the live table's `TimePartitioning.Field` / `Clustering.Fields` against the configured `partitionBy` / `clusterBy` (case-insensitive; range-partitioned tables always mismatch since ingestr never creates them).


iremcaginyurtturk:
Superseded by #970 — recreates the table via a safer **rename-aside** (rename target aside → swap into a fresh target → drop old, restoring on failure) instead of the plain drop-then-recreate. Live-verified for DATE and TIMESTAMP partition columns on both the copy-job and CTAS swap paths.

PR #950, opened 2026-07-09
cdc: deprecate the ?mode= URI parameter

turtleDev:
## What

`?mode=` and `--stream` both selected between a bounded and a continuous read, but only `--stream` was ever complete: it swaps the pipeline's write path for `StreamingExecutor`, which flushes on an interval or record count and commits the source's durable position after each flush. `mode=` only removed the stop condition from the source's internal read loop.

The two overlapped badly:

- Postgres, SQL Server and MongoDB CDC all **overwrote** the URI's mode when `--stream` was set, so `mode=stream` added nothing.
- Set **on its own**, `mode=stream` was worse than useless: the source read forever while the ordinary batch write path waited for a read that never ended. Nothing was written to the destination and no position was ever committed. For Postgres the replication slot froze too, since idle commit tokens are only emitted when streaming. Nothing validated against this.

That left `mode=` with two values — `batch`, already the default everywhere, and `stream`, either redundant or broken. The parameter carried no information.

## Changes

**Deprecate `mode=`, validated centrally.** `IngestConfig.validateCDCMode()` handles every `+cdc` source in one place, before any connection is attempted:

| URI | Result |
| --- | --- |
| `mode=stream` without `--stream` | rejected: *mode=stream is no longer supported; pass --stream to ingest continuously* |
| `mode=stream` with `--stream` | tolerated, deprecation warning |
| `mode=batch` | no-op, deprecation warning |
| `mode=once` | rejected: *invalid mode* |

Existing streaming and batch URIs keep working.

**Remove `CDCMode` from all four CDC sources**, so `opts.Streaming` is the only input deciding whether a read loop is bounded. `mode == ModeBatch` becomes `!opts.Streaming` at each stop-condition site; `streamLoop` loses its `mode` parameter, `applyStreamingMode` is deleted, and MongoDB's `streamCollection` loses its `mode` argument. In MySQL the `Mode` field was always `batch`, so its three stop conditions collapse to their right-hand clauses. Each source still strips `mode=` from the driver connection string.

This also closes the hang at the source layer: the multi-table Postgres reader required `Mode == ModeBatch && !opts.Streaming` before computing a target LSN, which is precisely why `mode=stream` alone left `targetLSN` at zero and never exited.

## Notes

- Parse-time mode tests moved out of the source packages into a table-driven `TestIngestConfigValidate_CDCMode`, which also covers case-insensitivity and `mode=` on a non-CDC source.
- Integration tests still pass `mode=batch` in their URIs and are unchanged — that is the backwards-compatibility guarantee exercised.
- Docs: `mode` documented as deprecated in `postgres.md`, dropped from the `mysql.md` and `vitess.md` examples.

### DIFF
--- a/.claude/skills/bigquery-destination/SKILL.md
+++ b/.claude/skills/bigquery-destination/SKILL.md
@@ -1,0 +1,177 @@
+---
+name: bigquery-destination
+description: How the ingestr BigQuery destination works — the staging→swap write path, load methods (load_job/storage_write), where dedup happens, copy-job vs CTAS swap selection, partition/cluster change handling (rename-aside), and merge partition pruning. Use when changing anything in pkg/destination/bigquery/ or BigQuery-affecting behavior in the replace/merge strategies (pkg/strategy/replace.go, pkg/strategy/merge.go). If your change alters behavior described here, update this file in the same change.
+---
+
+# BigQuery Destination — Implementation
+
+How ingestr writes data to BigQuery, including how partition/cluster changes are handled.
+
+> **Maintenance:** this file documents current behavior. If your change alters anything described here, update the relevant section in the same PR.
+
+## 1. Write Path
+
+Data flows through a staging table:
+
+```
+source  ->  staging table  ->  destination table
+```
+
+- The staging table is created first with the configured partition/clustering.
+- Data is loaded into staging.
+- Staging is then swapped into the destination.
+
+## 2. Loading Data into Staging (load methods)
+
+- **`load_job`** (default): a BigQuery load job.
+- **`storage_write`**: the BigQuery Storage Write API.
+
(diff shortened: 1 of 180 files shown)


turtleDev:
Good catch — fixed in d67d5cb8.

You are right that `mysql+cdc` has no `SupportsStreaming()`, so the old wording routed those users straight into a second error from the pipeline's capability gate. The message now states what is wrong with the parameter and leaves the capability question to the source:

> mode=stream is no longer supported; remove it and use --stream on sources that support continuous ingestion

I deliberately did not restore a MySQL-specific message. `validateCDCMode` runs before `Connect`, so it has no way to know whether the source implements `StreamingSource` without reaching into the registry — and the pipeline already reports that accurately at `pkg/pipeline/pipeline.go:87`.

iremcaginyurtturk (approved):

PR #951, opened 2026-07-09
Harden PostgreSQL CDC state, TRUNCATE, and stress coverage

karakanb:
## What
- move PostgreSQL CDC durability state into shared destination-managed tables: `_bruin_staging.cdc_state` and `_bruin_staging.cdc_targets`
- isolate multiple connectors with stable connector/source identities, fenced generations, destination incarnations, schema fingerprints, leases, target ownership, and v1/v2 adoption
- make snapshot/WAL handoff, checkpoint advancement, transaction boundaries, TRUNCATE, TOAST partial updates, slot replacement, and crash recovery fail closed
- keep dynamically discovered tables side-effect free until a supervised restart can admit them through the normal snapshot path
- add destination-specific state, claim, conditional truncate, identifier, and CDC merge handling across the supported SQL/warehouse destinations
- expand PostgreSQL CDC stress coverage for late tables, add/drop/rename/type DDL, PK updates, JSONB/TOAST, deletes, transactional TRUNCATE, restarts, and exact convergence

## Why
Resume previously depended on user-table LSNs and early state designs split markers across connector/table-specific tables. That could admit incomplete snapshots, collide when connectors shared a destination, and leave ambiguous recovery behavior after crashes, destination recreation, slot loss, or schema changes.

The shared append-only event model makes a checkpoint resumable only when its run fence, per-table snapshot completion, destination incarnation, and schema identity agree. The target registry prevents independent connectors from silently owning the same physical table.

## Impact
Eligible destinations create one shared `cdc_state` event table and one shared `cdc_targets` registry in their managed staging namespace. Connector IDs isolate state for different sources or connector instances. Losing the state table intentionally fails closed when target rows still exist; an explicit full refresh is required to rebuild safely.

A running PostgreSQL CDC stream that discovers a new eligible table exits before destination side effects. Its supervisor restart admits and snapshots the table through the normal startup path.

## Validation
- `make format`
- `make lint`
- `make test` (race-enabled Go suite plus Python SDK tests)
- `make test-conformance`
- focused PostgreSQL CDC integration, conformance, lease, crash, state-loss, restart, naming, TOAST, and destination protocol tests
- complex-workload stress test: 158,684 effective operations, 882 ops/sec, eight schema phases, two safe late-table restarts, zero operation errors, aggregate and row-by-row convergence
- live PostgreSQL CDC to BigQuery in `bruin-ingestr-playground`: initial snapshot, 18 KB TOAST value, shared state/target tables, checkpoint resume, update/delete/insert accuracy, and LSN advancement
- manual local PostgreSQL, SQLite, MySQL, Oracle, MSSQL, DuckDB, and CrateDB destination coverage through focused and full conformance runs


karakanb:
@greptileai

albertobruin (dismissed):

karakanb:
@greptile-apps

karakanb:
@greptile-apps

karakanb:
@greptile-apps Please review the final head c047b436 after the CI-only dead-code cleanup. Keep the review scoped to correctness issues in the existing implementation.

DjamilaBaroudi (approved):

PR #958, opened 2026-07-11
Add incremental predicate support for BigQuery merge loads

karakanb:
Adds an `--incremental-predicate` flag that lets you AND an extra condition into the MERGE ON clause on BigQuery, using `t`/`s` aliases for target and staging. This limits how much of the target table gets scanned during merges, which cuts costs significantly on large partitioned tables. The predicate is validated up front: it only works with the merge strategy on BigQuery, single-table mode, and not with `--full-refresh`. Also plumbed through the server API and the Python runner, with docs and tests.


karakanb:
@greptileai review

turtleDev (approved):

PR #1005, opened 2026-07-20
Speed up SQL Server replace bulk loads

karakanb:
## Summary

- bulk-load known-unique SQL Server replacement data into an unindexed staging heap, then create its primary key atomically during the swap
- use 12 staging writers for that fast path while preserving explicit concurrency settings
- default SQL Server connections to 32 KB TDS packets while preserving user overrides

## Benchmark

PostgreSQL → SQL Server, 1M rows, one warmup and five measured runs:

- before: 22.565s ± 1.075s
- after: 4.384s ± 0.113s
- 5.15× faster (44k → 228k rows/s)

## Validation

- `make format`
- `make lint`
- `make test`
- remote five-run 1M benchmark

### DIFF
--- a/pkg/destination/destination.go
+++ b/pkg/destination/destination.go
@@ -20,6 +20,7 @@ type PrepareOptions struct {
 	Schema                 *schema.TableSchema
 	DropFirst              bool
 	PrimaryKeys            []string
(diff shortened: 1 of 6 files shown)


karakanb:
@greptileai review

iremcaginyurtturk (approved):

PR #1046, opened 2026-07-31
fix(bigquery): cap query-job start retries so a persistent rate limit fails fast

TanayBensuYurtturk:
## Problem
`startQueryJobWithRetry` retried the BigQuery job-submission call (`query.Run`) in an **unbounded loop** on any retryable error. When the error was persistent — e.g. a sustained `rateLimitExceeded` from issuing more than the allowed DDL/metadata operations in a short window (>8 `ALTER COLUMN` in ~10s) — the loop spun **forever**:
- it never succeeded (submission kept getting rate-limited),
- it never failed (`rateLimitExceeded` is classified retryable, so `!retry` never triggered),
- it only logged at `config.Debug`, so with `--debug` off it looked like a **silent hang** with no feedback.

## Fix
- Cap the job-start retries at `queryJobStartMaxAttempts` (5). On exhaustion, return a clear error (`failed to start <op> query job <id> after N attempts: …`) while still resolving the CDC job fence, so a persistently retryable condition **fails cleanly** instead of looping forever.
- Surface each retry via `output.Warnf` instead of `config.Debug`, so the retrying is **visible** (no more silent hang).

Scope is limited to BigQuery's job-submission loop; no other destination has this pattern (they run queries synchronously via `ExecContext`). The backoff-delay computation (`min(attempt, queryJobMaxAttempts)`) is unchanged.

### DIFF
--- a/pkg/destination/bigquery/bigquery.go
+++ b/pkg/destination/bigquery/bigquery.go
@@ -22,6 +22,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/bruin-data/ingestr/internal/annotation"
 	"github.com/bruin-data/ingestr/internal/config"
+	"github.com/bruin-data/ingestr/internal/output"
 	"github.com/bruin-data/ingestr/pkg/destination"
 	"github.com/bruin-data/ingestr/pkg/schema"
 	"github.com/bruin-data/ingestr/pkg/source"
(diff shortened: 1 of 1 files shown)


TanayBensuYurtturk:
Addressed in e159711c. On start-retry exhaustion we now route through `reconcileAmbiguousBigQueryJob` instead of blindly resolving the fence: since `query.Run` uses a stable job ID, a retryable error can come back after BigQuery accepted the job, so reconcile cancels/confirms an accepted job (no orphan) and resolves the fence in every path. A `(nil, nil)` reconcile result means the job never landed (e.g. a persistent `rateLimitExceeded`), in which case we surface a clear `failed to start … after N attempts` error.

mustafa3rsan (approved):

PR #1090, opened 2026-08-12
Bound source record batches by bytes via --batch-size (Fluxx first)

iremcaginyurtturk:
## Problem

Sources cut a new record batch every `page-size` rows (default 25,000) with **no byte limit**, so wide-row tables build one enormous in-memory batch (e.g. Fluxx `client_store`: ~460 MB) that can exhaust memory. The OOM happens **at batch construction**, so it can only be bounded there — in the source's accumulation loop, before the giant batch is built.

## Change

- **`--batch-size`** flag (MiB, default 256; env `INGESTR_BATCH_SIZE`) → `config.MaxBatchBytes` → `ReadOptions.MaxBatchBytes`.
- **Fluxx** flushes when `rows >= page-size` **OR** the accumulated rows' estimate size `>= MaxBatchBytes`


cmd/ingest.go
```diff
@@ -124,6 +124,12 @@ func IngestCommand() *cli.Command {
 				Value:   25000,
 				Sources: cli.EnvVars("PAGE_SIZE", "INGESTR_PAGE_SIZE"),
 			},
+			&cli.IntFlag{
+				Name:    "batch-size",
+				Usage:   "Max size of a source record batch in MiB; sources flush on this or page-size, whichever comes first (0 = unlimited)",
```
turtleDev: Rather than having 0 as an escape hatch, always consider this as an upper bound on batch. Let users specify a larger batch size according to their needs.

pkg/source/fluxx/fluxx.go
```diff
@@ -468,17 +469,23 @@ func (s *FluxxSource) readResource(ctx context.Context, resourceName, endpoint,
...
+				normalized := normalizeFluxxItem(itemMap, fieldsToExtract)
+				allItems = append(allItems, normalized)
+				// Flush on bytes too, not just row count, so wide rows don't balloon.
+				if opts.MaxBatchBytes > 0 {
+					accBytes += arrowconv.RowBytes(normalized)
+				}
 			}
 
-			if len(allItems) >= batchSize {
+			if len(allItems) >= batchSize || (opts.MaxBatchBytes > 0 && accBytes >= opts.MaxBatchBytes) {
```
turtleDev: This can still produce a larger batch than the limit. We should slice the batch like we did in the downstream code.

turtleDev (dismissed):

turtleDev:
Looks good, but I have 2 major concerns:

* The overhead of serialising json payloads applies to each row. This can add up quickly and significantly affect our performance.
* These changes seems non-generalisable. This means it will be a redundant addition to every source implementation. This is usually a warning signal that there's a better approach/design that we should pursue.

turtleDev (approved):

Issue #806, opened 2026-06-16
MSSQL custom query (query:) drops DECIMAL/NUMERIC scale -> values rounded to integer (mapper defaults scale=0; DecimalSize() not used)


krom-als:
## Summary

For a Microsoft SQL Server **custom query** source (`--source-table "query:..."`), ingestr loads every `DECIMAL`/`NUMERIC` column with **scale 0**, rounding all values to integers. The same column read via a **native** `--source-table schema.table` is loaded with correct scale. No error is raised — silent numeric corruption.

## Environment

- ingestr **v1.0.33**
- Source: **Microsoft SQL Server** (go-mssqldb); column type `decimal(21,8)`
- Destination: Snowflake (also reproduces on other destinations)

## Reproduction

Given a table `my_table` with column `amount decimal(21,8)` holding e.g. `19.95000000`, `1722.76000000`:

**Native table source — correct:**
```
ingestr ingest --source-table "<db>.<schema>.my_table" ...
```
→ `AMOUNT` lands as `NUMBER(21,8)`, value `19.95000000` ✅

**Custom query source — rounded:**
```
ingestr ingest --source-table "query:SELECT amount FROM <db>.<schema>.my_table" ...
```
→ `AMOUNT` lands rounded to `20` (and `1722.76 → 1723`) ❌

Casting inside the query confirms only the type name matters: `CAST(amount AS DECIMAL(38,8))` → still rounded; `CAST(amount AS FLOAT)` → preserved; `CAST(amount AS VARCHAR)` → preserved.

## Root cause

The custom-query path types result columns from the bare driver type name, which has no precision/scale, and the mapper then defaults DECIMAL/NUMERIC to **scale 0**.

`pkg/source/mssql/mssql.go` (custom-query reader, ~L488):
```go
colTypes, err := rows.ColumnTypes()
...
for i, ct := range colTypes {
    dt, precision, scale, arrayType := MapMSSQLToDataType(ct.DatabaseTypeName())  // "DECIMAL" — no (p,s)
    ...
    columns[i] = schema.Column{..., Precision: precision, Scale: scale, ...}
}
```
`ct.DatabaseTypeName()` returns `"DECIMAL"` (no parameters), so in `pkg/source/mssql/mapper.go` (L56-57):
```go
case "DECIMAL", "NUMERIC":
    return schema.TypeDecimal, 18, 0, schema.TypeUnknown   // scale 0 -> rounds
```
[...]

PR #1124, opened 2026-08-27
fix(clickhouse): read decimal precision and scale in GetTableSchema

steache:
## What

The ClickHouse destination never reads a decimal column's precision and scale back from the
server, so schema evolution compares against a phantom type and **every sync after the first
fails** for any table containing a decimal column.

`ClickHouseDestination.GetTableSchema` builds each column as `{Name, DataType, Nullable}` and
leaves `Precision`/`Scale` at zero. `schemaevolution.normalizedDecimal` then fills
`Precision = 38` but leaves `Scale = 0`, producing a phantom `Decimal(38, 0)`. Widening a real
`Decimal(38, 9)` source column against that phantom computes integer digits `38 - 0 = 38`,
scale `max(9, 0) = 9`, and asks for precision `38 + 9 = 47`:

```
schema evolution failed: failed to compare schemas: column <c>:
decimal widening requires precision 47 (integer digits 38 + scale 9),
maximum supported precision is 38
```

The failure cannot occur on a first load — there is no destination table to compare against —
so initial testing always passes and the error only appears on the next run. Neither
`--schema-contract` (`freeze`, `discard_value`, `discard_row`) nor casting the source column
avoids it: the schema comparison runs before the contract is consulted, and the source side was
never the broken one.

## Changes

- **`pkg/destination/clickhouse/clickhouse.go`**
  - `GetTableSchema` now populates `Precision`/`Scale` for `schema.TypeDecimal` columns.
  - New `parseClickHouseDecimal` handles both renderings plus wrappers:
    - `Decimal(P, S)` — `DESCRIBE` emits a **space** after the comma;
    - `Decimal32|64|128|256(S)` — precision implied by width (9 / 18 / 38 / 76), which is what
      ingestr's own type mapper emits (`Decimal128(9)` by default);
    - `Nullable(...)` and `LowCardinality(...)`, unwrapped recursively.
  - Non-decimal types return `0, 0` — exactly what they returned before.

- **`pkg/destination/clickhouse/clickhouse_test.go`**
  - Table test covering both renderings, both wrappers and non-decimal types.
  - `TestParseClickHouseDecimalPreservesScaleForWidening` — named regression test pinning the
    precision-47 failure.

No docs change: the ClickHouse **destination** has no docs page (only
`docs/supported-sources/clickhouse.md`), and this fixes existing behaviour rather than adding an
option. No `go.mod`/`go.sum` change, so the license audit lock is untouched.

## How to test

1. `go test ./pkg/destination/clickhouse/`
2. End to end: ingest a source table with a `Decimal(38, 9)` column into ClickHouse **twice**.
   Before this change the second run fails with the error above. After it, the destination reads
   back as `(38, 9)`, the comparison short-circuits, and no `ALTER` is emitted at all.

Verified end to end against a live multi-replica ClickHouse over 60+ tables and several million
rows: tables that previously failed on their second pass now complete, with
`count() FINAL == count()` on every table, so the merge path introduces no duplication.

## Risk & rollback

- **Risk:** low. The two fields being populated were previously always zero, and that zero is
  precisely what produced the failure — no correct behaviour can depend on it. Non-decimal
  columns are untouched, and there are no API or signature changes.

## Checklist

- [x] Unit tests added or updated (`make test`, `make test-full`)
- [x] Format and lint pass (`make format-ci` clean; `make lint` reports 0 issues on the
      changed package)
- [x] Integration tests pass if affected — ran the destination conformance suites that
      exercise this path (`TestDestinations_SchemaEvolution`, `TestDestinations_Merge`)
- [x] Tested locally

## Security

- [x] No secrets, credentials, or PII in this change
- [x] No new dependencies with known vulnerabilities
- [x] Security implications considered (if applicable) — none; the change is type-metadata
      parsing of a `DESCRIBE` result.

> Note for reviewers: `make lint-full` currently reports 3 pre-existing `staticcheck`
> SA1019 deprecation findings on `main` — `pkg/destination/snowflake/snowflake.go:300`,
> `pkg/source/googleads/googleads.go:63` and `pkg/source/sqs/sqs.go:384`. None are touched
> by this PR; `golangci-lint run ./pkg/destination/clickhouse/...` is clean.

## Related issues

None found. Searched the repository for decimal/precision and ClickHouse schema-evolution
issues; #806 (MSSQL custom query drops decimal scale), #940 (preserve Snowflake decimal
metadata) and #641 (Snowflake decimal handling) are adjacent but none covers the ClickHouse
read-back path.

### DIFF
--- a/pkg/destination/clickhouse/clickhouse.go
+++ b/pkg/destination/clickhouse/clickhouse.go
@@ -645,6 +645,9 @@ func (d *ClickHouseDestination) GetTableSchema(ctx context.Context, table string
 			DataType: mapClickHouseTypeToSchema(colType),
 			Nullable: strings.HasPrefix(colType, "Nullable("),
 		}
+		if col.DataType == schema.TypeDecimal {
+			col.Precision, col.Scale = parseClickHouseDecimal(colType)
+		}
 
 		columns = append(columns, col)
 	}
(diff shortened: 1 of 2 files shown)


iremcaginyurtturk:
Thanks for catching this and the fix! We’re going with #1167, which fixes the same bug in a more general way, so I’ll close this one. 🙏

PR #1128, opened 2026-08-28
feat(source): add Sklik source

steache:
## What

Adds [Sklik](https://www.sklik.cz/) as a source. Sklik is the paid-search advertising
platform of Seznam.cz, the dominant search engine in Czechia — for Czech advertisers it
sits alongside Google Ads rather than behind it, and there is currently no way to get its
spend into a warehouse through ingestr.

The connector speaks Sklik's JSON-RPC API and exposes five entity snapshots plus two
report tables.

## Changes

- **`pkg/source/sklik/sklik.go`** — the source. Handles Sklik's session model
  (`client.loginByToken` returns a *rolling* session that changes on every response,
  including re-login when one expires mid-run), the `*.list` methods, and the
  asynchronous `createReport` / `readReport` pair.
- **`pkg/source/sklik/register.go`** — registers the `sklik` scheme.
- **`pkg/source/sklik/sklik_test.go`** — covers report flattening, the account stamp,
  and a lock-in test for the primary-key choice.
- **`docs/supported-sources/sklik.md`**, **`docs/.vitepress/config.mjs`**,
  **`docs/supported-sources/platforms.md`** — docs page and both indexes.

### Tables

| Table | Primary key | Incremental | Strategy |
|---|---|---|---|
| `campaigns`, `groups`, `ads`, `keywords`, `conversions` | `id` | — | merge |
| `campaign_stats_daily` | `id`, `date` | `date` | merge |
| `search_queries` | `query`, `keyword_id`, `date` | `date` | merge |

`search_queries` keys on `keyword_id` as well as `(query, date)` because Sklik reports
the same search term once per matching keyword; without it the merge keeps one arbitrary
keyword's row per query and silently drops the rest of the spend.

### Three Sklik API behaviours worth knowing about

These are the reason several parts of the code look defensive, so flagging them for review:

1. **Throttling degrades to an empty HTTP 200, not a 429.** Sklik rate-limits per account
   per minute, and over the limit a report returns
   `{"status":200,"statusMessage":"OK","report":[]}`. A throttled report and a genuinely
   empty one are byte-identical, so a run can quietly load nothing and look healthy. The
   only reliable signal is `createReport`'s `totalCount`: the source retries with a linear
   backoff while it promises rows the read has not produced, and the retry loop only
   engages in that case — a healthy report never sleeps.
2. **`queries.createReport` returns empty unless scoped.** Without a campaign/group/keyword
   restriction it yields `report: []` with `status: 200`, which reads as "this account has
   no search terms". On a live account: unrestricted → 0 rows; restricted to one campaign
   over the same window → 165. `campaigns.createReport` does not behave this way, so a
   working campaign report proves nothing about the search-term one.
3. **`displayColumns` enums are per-method and asymmetric.** `ctr` is accepted by
   `campaigns.readReport` and rejected by `queries.readReport`, and Sklik answers with a
   bare `400 Bad arguments` for the whole report with no field-level detail. `ctr` is
   therefore deliberately absent from the search-query columns.

All three are documented in the docs page and in comments at the relevant code.

## How to test

```bash
ingestr ingest \
  --source-uri 'sklik://?token=<api_token>' \
  --source-table campaigns \
  --dest-uri duckdb:///sklik.duckdb \
  --dest-table main.campaigns
```

Report tables take `--interval-start` / `--interval-end`. A Sklik account with an API
token is needed; the token is generated in the Sklik UI.

`go test ./pkg/source/sklik/` covers the pure logic without network access.

## Risk & rollback

- **Risk:** low. Purely additive — one new package, one catalog line, three docs files.
  No existing source, destination or shared code path is touched, so nothing can regress
  for current users.

## Checklist

- [x] Integration tests pass if affected — not affected; no shared code path is touched
- [x] Tested locally against a live Sklik account

## Security

- [x] No secrets, credentials, or PII in this change — the token is supplied via the URI
      and never logged

## Related issues

None found.

### DIFF
--- a/docs/.vitepress/config.mjs
+++ b/docs/.vitepress/config.mjs
@@ -267,6 +267,7 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
               { text: "SFTP", link: "/supported-sources/sftp.md"},
               { text: "SharePoint", link: "/supported-sources/sharepoint.md" },
               { text: "Shopify", link: "/supported-sources/shopify.md" },
+              { text: "Sklik", link: "/supported-sources/sklik.md" },
               { text: "Slack", link: "/supported-sources/slack.md" },
               { text: "Smartsheet", link: "/supported-sources/smartsheets.md" },
               { text: "Snapchat Ads", link: "/supported-sources/snapchat-ads.md" },
--- a/docs/supported-sources/platforms.md
+++ b/docs/supported-sources/platforms.md
@@ -19,6 +19,7 @@ Use this page to browse platforms by category. The sidebar keeps the full alphab
 - [Mailchimp](/supported-sources/mailchimp.md)
 - [Pinterest](/supported-sources/pinterest.md)
 - [Reddit Ads](/supported-sources/reddit_ads.md)
+- [Sklik](/supported-sources/sklik.md)
 - [Snapchat Ads](/supported-sources/snapchat-ads.md)
 - [TikTok Ads](/supported-sources/tiktok-ads.md)
 
--- a/docs/supported-sources/sklik.md
+++ b/docs/supported-sources/sklik.md
@@ -1,0 +1,109 @@
+# Sklik
+
+[Sklik](https://www.sklik.cz/) is the paid-search advertising platform of Seznam.cz, the dominant Czech search engine.
+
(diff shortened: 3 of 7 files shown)


steache:
(used at our company for few weeks already with real credentials, so as long as we keep using Sklik, we will use it & push updates - but do not feel pressure to merge it, we can keep it in our fork :) )

iremcaginyurtturk (approved):

PR #1129, opened 2026-08-28
feat(source): add 2Checkout source

steache:
## What

Adds [2Checkout](https://www.2checkout.com/) (Verifone) as a source, via the
[REST API 6.0](https://verifone.cloud/docs/2checkout/API-Integration). 2Checkout is a
payments and subscription-billing platform widely used by SaaS and digital-goods
vendors, and has no ingestr source today.

## Changes

- **`pkg/source/twocheckout/twocheckout.go`** — the source. Handles 2Checkout's custom
  HMAC-SHA256 signing (`X-Avangate-Authentication`), pagination, and the per-endpoint
  filter differences described below.
- **`pkg/source/twocheckout/register.go`** — registers the `twocheckout` scheme.
  (`twocheckout`, not `2checkout`: RFC 3986 schemes must begin with a letter, and
  `url.Parse` rejects a leading digit.)
- **`pkg/source/twocheckout/twocheckout_test.go`** — covers the declared-column floor
  and the gross correction.
- **`docs/supported-sources/twocheckout.md`**, **`docs/.vitepress/config.mjs`**,
  **`docs/supported-sources/platforms.md`** — docs page and both indexes.

### Tables

| Table | Primary key | Incremental | Strategy |
|---|---|---|---|
| `orders` | `RefNo`, `Status` | `StartDate` / `EndDate` | merge |
| `subscriptions` | `SubscriptionReference` | `ModifiedAfter` | merge |
| `products` | `ProductCode` | — | merge |
| `promotions` | `code` | — | merge |

There is deliberately no `customers` table: REST 6.0 exposes only `/customers/search/`,
which requires an `Email` parameter, so the customer base cannot be enumerated.

### Three behaviours worth reviewer attention

1. **`GrossPrice` is served as a running cumulative total.** On the `/orders/` list
   endpoint, each order's `GrossPrice` / `GrossDiscountedPrice` is its own gross **plus
   every gross before it in the page**. This looks like an upstream serialisation bug.
   `NetPrice`, `NetDiscountedPrice` and `VAT` are correct and `gross == net + vat` holds
   for every order (including refunds, where all three are negative), so the source
   recomputes gross rather than ingesting the raw field — loading it as served would put
   silently, plausibly wrong revenue in the destination. The correction is guarded by an
   epsilon comparison, so it is a no-op on a correct payload and disappears by itself if
    2Checkout fixes the serialisation.

2. **`orders` is keyed on `(RefNo, Status)`, not `RefNo`.** An order moves between
   statuses over its life (`COMPLETE` → `REFUND` → `REVERSED`) and 2Checkout negates the
   money fields on a refund row. Keyed on `RefNo` alone, `merge` overwrites the original
   charge with the refund and only negative amounts remain. Keying on the pair keeps both
   rows. It does not recover history — `/orders/` returns current state only — but it
   preserves transitions from the first load onward.

3. **`/subscriptions/` honours only `ModifiedAfter`.** It refuses to page past 20,000
   results, so an unfiltered pull of a large account is impossible rather than slow, and
   `StartDate`, `EndDate` and `ProductCode` are silently ignored (the returned `Count` is
   identical with and without them). `ModifiedAfter` is the right incremental key for
   `merge` since it catches changed records, but it means a narrow window returns
   recently-*touched* rather than recently-*created* subscriptions.

A smaller one: `orders` declares a minimum column set so that fields which are null in
every row of a batch (`Source`, `ExternalReference`, `ApproveStatus`) still reach the
destination. Schema inference otherwise drops such a column — it survives arrowconv but
lands as `UnknownArrowType` and is dropped later when `KnownSchema` is false. Declared
columns are additive, so every other key the API returns still passes through.

## How to test

```bash
ingestr ingest \
  --source-uri 'twocheckout://?merchant_code=<code>&secret_key=<key>' \
  --source-table orders \
  --dest-uri duckdb:///twocheckout.duckdb \
  --dest-table main.orders
```

A 2Checkout merchant account with API credentials is needed. Note the signature embeds a
timestamp, so an inaccurate host clock produces a `401` that reads like a bad credential;
the source says so in the error message.

`go test ./pkg/source/twocheckout/` covers the pure logic without network access.

## Risk & rollback

- **Risk:** low. Purely additive — one new package, one catalog line, three docs files.
  No existing source, destination or shared code path is touched.

## Checklist

- [x] Integration tests pass if affected — not affected; no shared code path is touched
- [x] Tested locally against a live 2Checkout account

## Security

- [x] No secrets, credentials, or PII in this change — credentials are supplied via the
      URI and never logged

## Related issues

None found.

### DIFF
--- a/docs/.vitepress/config.mjs
+++ b/docs/.vitepress/config.mjs
@@ -277,6 +277,7 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
               { text: "TikTok Ads", link: "/supported-sources/tiktok-ads.md" },
               { text: "Trello", link: "/supported-sources/trello.md" },
               { text: "Twilio", link: "/supported-sources/twilio.md" },
+              { text: "2Checkout", link: "/supported-sources/twocheckout.md" },
               { text: "Typeform", link: "/supported-sources/typeform.md" },
               { text: "Wise", link: "/supported-sources/wise.md" },
               { text: "Wistia", link: "/supported-sources/wistia.md" },
--- a/docs/supported-sources/platforms.md
+++ b/docs/supported-sources/platforms.md
@@ -47,6 +47,7 @@ Use this page to browse platforms by category. The sidebar keeps the full alphab
 
 ## Commerce, payments, billing, and finance
 
+- [2Checkout](/supported-sources/twocheckout.md)
 - [Apple App Store](/supported-sources/appstore.md)
 - [Chargebee](/supported-sources/chargebee.md)
 - [FastSpring](/supported-sources/fastspring.md)
--- a/docs/supported-sources/twocheckout.md
+++ b/docs/supported-sources/twocheckout.md
@@ -1,0 +1,71 @@
+# 2Checkout
+
+[2Checkout](https://www.2checkout.com/) (Verifone) is a payments and subscription-billing platform for digital goods and SaaS.
+
+ingestr supports 2Checkout as a source through the [2Checkout REST API 6.0](https://verifone.cloud/docs/2checkout/API-Integration).
+
+## URI format
+
+```plaintext
+twocheckout://?merchant_code=<merchant_code>&secret_key=<secret_key>
+```
+
+URI parameters:
+
+- `merchant_code`: Required. The merchant code from the 2Checkout Control Panel.
+- `secret_key`: Required. The API secret key issued alongside it.
+
+The scheme is `twocheckout`, not `2checkout`: RFC 3986 URI schemes must begin with a letter.
+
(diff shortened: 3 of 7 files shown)


karakanb:
@steache hey, thanks a lot for your contributions!

Two questions:
- did you implement these sources because you need them or simply as you thought this could be interesting to someone else?
- did you test these with real credentials yourself?

steache:
@karakanb thanks for fast response! :)

- yes, we use them daily for few weeks already :)
- yes

(dont feel pressure to merge it, closing it is absolutely OK - and I will just keep it in our fork, just wanted to give a bit back)

karakanb:
@steache sounds awesome, we'll look into this and get it going quickly. cheers!

iremcaginyurtturk (approved):

PR #1151, opened 2026-08-28
feat(source): add Exchangerates API source

steache:
## What

Adds [exchangeratesapi.io](https://exchangeratesapi.io/) (APILayer) as a foreign-exchange
source.

This complements the existing `frankfurter` source rather than replacing it. frankfurter
serves free ECB reference rates: ~30 currencies, EUR base, and **no weekend rows**,
because the ECB does not publish on weekends. That remains the better default. A paid
exchangeratesapi.io plan publishes weekends, ~172 currencies and a switchable base.

**On the free tier this source is pointless** — that tier *is* ECB data, EUR base, so it
returns numbers identical to frankfurter but with an API key to rotate. The docs page says
so plainly.

## Changes

- **`pkg/source/exchangeratesapi/`** — source, `register.go`, tests.
- Docs page and both indexes.

### Two behaviours worth reviewer attention

1. **Historical rates are not reproducible.** The API answers with what it believes today,
   so asking again later gives a different answer for the same past date. Across 16 years
   of stored rates, two collectors overlapping on 32,089 (date, currency) pairs disagreed
   on **31,878 of them (99.3%)**, by up to 10.1%. Stored rates are a record of what was
   quoted at the time and can be preserved but never regenerated — worth knowing before
   anyone uses this to "rebuild" history.
2. **`/timeseries` is plan-gated.** It returns `403 function_access_restricted` on plans
   where the single-date `/v1/{date}` endpoint works fine, so the source walks the interval
   one day at a time by design rather than by oversight. On a plan that includes it,
   switching is a contained change to one function.

The source also emits a base-to-base identity row (rate 1.0), matching frankfurter, so
converting an amount already in the base currency finds a row rather than yielding NULL.

## How to test

```bash
ingestr ingest \
  --source-uri 'exchangeratesapi://?api_key=<key>&base=EUR' \
  --source-table exchange_rates \
  --dest-uri duckdb:///fx.duckdb \
  --dest-table main.exchange_rates
```

## Risk & rollback

## Checklist

- [x] Tested locally against a live account

## Security

- [x] No secrets, credentials, or PII in this change

## Related issues

None found.

---

### Where this comes from


docs/supported-sources/exchangeratesapi.md
```diff
@@ -0,0 +1,86 @@
...
+> models join FX on an exact date rather than as-of, that gap becomes a zero rather than a
+> carried-forward rate.
+
+> [!CAUTION]
+> **Historical rates are not reproducible, so do not use this source to rebuild history.** The
+> API answers with what it believes today; asking again later returns a different answer for the
+> same past date. Measured on 16 years of stored rates from one production pipeline, two
+> collectors overlapping on 32,089 `(date, currency)` pairs disagreed on 31,878 of them (99.3%),
+> by up to 10.1%. Stored rates are a record of what was quoted at the time — preserve them, and
+> use this source to move forward.
```
karakanb, line 21: This is a super-interesting problem. Have you reported this to the API owners? Seems like this shouldn't have been the case, no? 

Could it be that depending on the time of the day you ran it it would report back the currency at that date at that exact time?

karakanb (dismissed):

docs/supported-sources/exchangeratesapi.md
steache: Not reported, its absolutely the case - it depends on the time of the day you asked first. But if you ask after few days/weeks/... you get one number for the rest of the time.

iremcaginyurtturk (approved):

PR #1153, opened 2026-08-28
feat(source): add Twenty CRM source

steache:
## What

Adds [Twenty](https://twenty.com/) as a source — an open-source CRM, available both hosted
(api.twenty.com) and self-hosted. Reads objects and records through its REST API.

## Changes

- **`pkg/source/twenty/`** — source, `schema.go`, `register.go`, tests.
- Docs page and both indexes.

### Metadata-derived schemas

The source has a fixed list of standard tables and supports workspace-specific objects as
`custom:<object_name>`. It reads each object's metadata at runtime because fields differ
between workspaces, including custom fields.

### Two decisions worth reviewer attention

1. **Composite field types use ingestr's JSON type** (`EMAILS`, `PHONES`, `LINKS`,
   `ADDRESS`, `CURRENCY`, `FULL_NAME`, `ACTOR`, `RICH_TEXT_V2`, `ARRAY`, `MULTI_SELECT`,
   `RAW_JSON`). `CURRENCY` amounts retain their exact integer micros representation.
2. **An object without an `id` is refused rather than ingested.** There would be nothing
   to deduplicate on, and under an append-style incremental strategy every run would add
   the whole object again.

Note Twenty soft-deletes: rows carry `deletedAt` and are returned, so a consumer wanting
the live set filters `deletedAt IS NULL`. The docs page says so.

## How to test

```bash
ingestr ingest \
  --source-uri 'twenty://api.twenty.com?api_key=<key>' \
  --source-table people \
  --dest-uri duckdb:///twenty.duckdb \
  --dest-table main.people
```

## Risk & rollback

## Checklist

- [x] Tested locally against two live workspaces

## Security

- [x] No secrets, credentials, or PII in this change — the API key comes from the URI and
      is never logged

## Related issues

None found.

---

### Where this comes from

### DIFF
--- a/docs/supported-sources/twenty.md
+++ b/docs/supported-sources/twenty.md
@@ -37,23 +37,30 @@ ingestr ingest \
 
 ## Tables
 
-The table name is an object's **plural API name** — `people`, `companies`,
-`opportunities`, `notes`, `tasks`, `leads`, `workspaceMembers`, and any custom
-object your workspace defines.
+Twenty supports the following tables:
 
-There is no fixed table list, because Twenty has no fixed object list. Every
-object, standard or custom, publishes its fields and types under
-`/rest/metadata/objects`, and ingestr reads that to build the table's schema. Two
-workspaces of the same Twenty version can therefore expose different tables and
-very different columns, and both are handled by the same connector.
+| Table | PK | Inc Key | Inc Strategy | Details |
+|-------|----|---------|--------------|---------|
+| `companies` | id | updatedAt | merge | Companies in the workspace. |
+| `notes` | id | updatedAt | merge | Notes attached to workspace records. |
+| `opportunities` | id | updatedAt | merge | Sales opportunities. |
+| `people` | id | updatedAt | merge | People in the workspace. |
+| `tasks` | id | updatedAt | merge | Tasks attached to workspace records. |
+| `workspaceMembers` | id | updatedAt | merge | Members of the workspace. |
+| `custom:<object_name>` | id | updatedAt | merge | A custom object, using its plural API name. |
 
-Passing a singular name is caught and corrected:
+Twenty exposes custom objects through the same REST API as standard objects. To
+ingest one, prefix its plural API name with `custom:`, for example
+`custom:leads`. The connector reads the object's metadata at runtime, so custom
+fields are included automatically.
 
+```sh
+ingestr ingest \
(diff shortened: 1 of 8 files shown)


pkg/source/twenty/schema.go
```diff
@@ -0,0 +1,310 @@
...
+
+// dataTypeFor maps a Twenty field type onto an ingestr/Arrow type.
+//
+// ⚠️ COMPOSITE FIELDS BECOME STRINGS CARRYING JSON TEXT, NOT schema.TypeJSON.
+// Twenty's composites (EMAILS, PHONES, LINKS, ADDRESS, CURRENCY, FULL_NAME,
+// ACTOR, RICH_TEXT_V2, ARRAY, MULTI_SELECT, RAW_JSON) are small JSON objects.
+// TypeJSON is not handled uniformly across destinations — some have
+// no mapping for it — so declaring it would land us on an untested path. String
+// is what the sibling `abra` source does with the same class of value, and a
+// reads it back with JSONExtractString, a pattern already in use here.
```
karakanb: I don't think this is correct. Can you please update it to use JSON instead?

karakanb (commented):
Thanks a lot for the PR!

Could you please address the JSON comment, and remove the emojis and the amount of comments from the code? it is clear to be AI-generated, which is fine for the code, but the comments seem to have quite a bit of redundancy in general.

karakanb:
@greptileai

docs/supported-sources/twenty.md
```diff
@@ -0,0 +1,109 @@
...
+
+The table name is an object's **plural API name** — `people`, `companies`,
+`opportunities`, `notes`, `tasks`, `leads`, `workspaceMembers`, and any custom
+object your workspace defines.
+
+There is no fixed table list, because Twenty has no fixed object list. Every
+object, standard or custom, publishes its fields and types under
+`/rest/metadata/objects`, and ingestr reads that to build the table's schema. Two
+workspaces of the same Twenty version can therefore expose different tables and
+very different columns, and both are handled by the same connector.
```
karakanb: It seems like there's a fixed table list already, so I would appreciate it if you could make the docs reflect the same structure as the other sources.

Also, can you please prefix the custom table names with `custom:` so that it looks like `custom:<your object name>`, similar to other sources we have.

karakanb (approved):

PR #1154, opened 2026-08-28
feat(source): add Fakturoid source

steache:
## What

Adds [Fakturoid](https://www.fakturoid.cz/) as a source — a Czech invoicing and accounting
service. Reads invoices (with their lines and VAT-rate summaries) and subjects through
API v3, authenticating with OAuth2 `client_credentials`.

## Changes

- **`pkg/source/fakturoid/`** — source, `fields.go`, `register.go`, tests.
- Docs page and both indexes.

### `user_agent` is a required parameter with no default

Fakturoid rejects a missing or generic `User-Agent` — it is a documented hard requirement,
not a courtesy — and it does so with a **403 on every endpoint including `/oauth/token`**,
so it reads like a credential problem rather than a header problem.

We deliberately did not ship a default: a shared default would send one user's contact
address on every other user's traffic. The error message explains the requirement.

### Three more behaviours worth reviewer attention

1. **Pagination is fixed at 40 rows with no total count.** `per_page` is not a parameter —
   the page size is a server constant, so the only end-of-data signal is a short page.
2. **Field lists are allow-lists, not "whatever the API returns".** Fakturoid returns
   fields beyond those projected (`attachments`, `eet_records`, `legacy_bank_details`,
   `vat_rates_summary` on the invoice itself); emitting everything would make the
   destination shape follow whatever the vendor adds next. Fields the API returns that are
   not projected are counted and logged once per run as drift rather than silently dropped.
3. **`merge` cannot observe a deletion.** A line removed from an existing invoice lingers
   in the destination — the API stops returning it with no tombstone. Invoice and subject
   deletions are equally invisible. Use a periodic full reload if deletions matter.

`invoices_lines` and `invoices_vat_rates` are derived from the same `/invoices.json`
payload, so requesting them re-pages the invoice list.

## How to test

```bash
ingestr ingest \
  --source-uri 'fakturoid://?client_id=<id>&client_secret=<secret>&slug=<slug>&user_agent=MyCompany%20(billing@mycompany.com)' \
  --source-table invoices \
  --dest-uri duckdb:///fakturoid.duckdb \
  --dest-table main.invoices
```

## Risk & rollback

## Checklist

- [x] Tested locally against a live account

## Security

- [x] No secrets, credentials, or PII in this change — credentials come from the URI and
      are never logged

## Related issues

None found.

---

### Where this comes from


iremcaginyurtturk:
Thanks for this, @steache. Nice connector, and your notes made it easy to review. I ran it against a live account and it worked great.

I pushed a commit on top with a few changes to match how ingestr sources usually work:

- Column types now come from schema inference (timestamps, dates, ints, booleans, JSON) instead of everything being text.
- All API fields pass through now, so payments, tags, attachments and friends aren’t dropped.
- Removed _etl_loaded_at (ingestr adds _ingestr_loaded_at already).

You’re not locked into any of this, though: --columns lets you override types (e.g. back to string) and pin columns so they always show up even when empty, and --exclude-columns drops what you don’t need.

Happy to change anything. Thanks again for contributing it!

iremcaginyurtturk (approved):

PR #1166, opened 2026-09-01
fix(clickhouse): read decimal precision/scale in GetTableSchema via shared mapper

iremcaginyurtturk:
## What

`ClickHouseDestination.GetTableSchema` left `Precision`/`Scale` unset, so schema evolution
compared a real `Decimal(38, 9)` source column against a phantom `Decimal(38, 0)` and **every
merge/append sync after the first failed** for any table with a decimal column:

```
decimal widening requires precision 47 (integer digits 38 + scale 9),
maximum supported precision is 38
```

## Fix

Read the schema from `system.columns` and reuse the source's existing `MapClickHouseToDataType`
instead of hand-parsing `DESCRIBE` output. This aligns the destination read-back with how the
ClickHouse source already reads schema, removes the duplicated local mapper, and populates
`Precision`, `Scale`, and `ArrayType`. `DateTime64(..., 'UTC')` is kept as `TypeTimestampTZ` so
timezone-aware timestamps round-trip.

`GetTableSchema` is used only for schema-evolution comparison and existence/naming detection; it
does not affect writes, table creation, or row data.

### DIFF
--- a/pkg/destination/clickhouse/clickhouse.go
+++ b/pkg/destination/clickhouse/clickhouse.go
@@ -22,6 +22,7 @@ import (
 	"github.com/bruin-data/ingestr/pkg/destination"
 	"github.com/bruin-data/ingestr/pkg/schema"
 	"github.com/bruin-data/ingestr/pkg/source"
+	srcclickhouse "github.com/bruin-data/ingestr/pkg/source/clickhouse"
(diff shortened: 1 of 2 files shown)


iremcaginyurtturk:
Superseded by #1167 (branch renamed to `fix/clickhouse-decimal-schema-readback`; the rename closed this one).

PR #1203, opened 2026-09-21
feat(hubspot): add HubSpot reverse-ETL destination

iremcaginyurtturk:
## What

Adds a **HubSpot destination** that writes CRM objects and associations from any ingestr source (reverse ETL).

## Strategies

| Strategy | Behavior | `--primary-key` |
|---|---|---|
| `merge` / `append` | Upsert or create records | required |
| `update` / `delete` | Address existing records; source column **and** `id_property` both default to `hs_object_id` when omitted | optional |
| `replace` | Full object-wide **mirror** — upsert everything in the source, archive everything not in it (including keyless records) | required |
| associations | `obj1+obj2` destination table links two object types | — |

## How matching works

Two axes:
- **Match property** — a HubSpot **internal name**, from `id_property=` on the dest table.
- **Source value column** — from `--primary-key`.

Property references (`id_property`, `--columns` dest side) are always HubSpot **internal names**, never display labels.

## Reliability

- HubSpot batch writes are **not atomic**, so a record-level `400`/`409` **bisects** the batch: valid rows still land while the offending record is isolated. Verified live at 560 rows with one poison row → 559 written, 1 rejected.
- Rejects honor `--reject-mode` (`fail` | `fail_fast` | `skip`); `skip` prints the offending records (named by identifier) at the end of the run.
- `resolveKeysToIDs` chunks `batch/read` to the 100-input API cap.
- `--write-nulls` (default: clear) applies only to property-writing strategies.

## Notes

- `--columns` on reverse-ETL errors when the named source column doesn't exist.
- Unresolved association keys become not-found rejects (honoring `--reject-mode`); empty cells are skipped.
- Docs at `docs/supported-sources/hubspot.md`.

## Testing

Unit tests cover all strategies, batch bisection, mirror (incl. keyless archive), association resolution, chunking, and PK/column-override resolution. Live-verified end-to-end against a HubSpot sandbox (all record/association strategies, mirror, batch bisection at scale).


iremcaginyurtturk:
@greptile review

terzioglub (approved):

